PR #11, opened 2020-06-23
Tweak horizontal example layout

bryankeller:
## Details
This PR fixes an assert I was hitting due to the horizontal calendar not having enough room to layout (without size going negative) on iPhone 8. The root cause was having a monthsWidth that was too constrained. This bug only affects the example project.

## Motivation and Context

Improve example app.

## How Has This Been Tested

Tested on a few different iPhone simulators and iPad simulator.

## Types of changes

## Checklist

### DIFF
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/DemoPickerViewController.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/DemoPickerViewController.swift
@@ -108,7 +108,7 @@ extension DemoPickerViewController: UITableViewDelegate {
     let demoViewController = demoDestinations[indexPath.item].destinationType.init(
       monthsLayout: monthsLayoutPicker.selectedSegmentIndex == 0
         ? .vertical(pinDaysOfWeekToTop: false)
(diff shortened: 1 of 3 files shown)


Example/HorizonCalendarExample/HorizonCalendarExample/DemoPickerViewController.swift
```diff
@@ -108,7 +108,7 @@ extension DemoPickerViewController: UITableViewDelegate {
     let demoViewController = demoDestinations[indexPath.item].destinationType.init(
       monthsLayout: monthsLayoutPicker.selectedSegmentIndex == 0
         ? .vertical(pinDaysOfWeekToTop: false)
-        : .horizontal(monthWidth: min(view.bounds.width - 128, 512)))
```
bryankeller, line 111: subtracting 128 from 375 just doesn't leave enough room for 7 days when there's horizontal spacing between individual days.

bryankeller: The good news is the internal assert works to let you know the layout isn't valid 😄

PR #14, opened 2020-06-25
Bk/partial month visibility

bryankeller:
## Details

This PR adds in partial boundary month visibility - an option available for vertical-layout calendars. 

ezgif-6-8a0d9f75b763

To do this, the `MonthsLayout.vertical` enum case has been changed so that it takes an instance of `VerticalMonthsLayoutOptions`. This new type holds the `pinDaysOfWeekToTop` and the newly introduced `alwaysShowCompleteBoundaryMonths` properties.

The existing case's definition has been marked as deprecated:
image

Because the public API has been expanded, and an existing API marked as deprecated, this warrants a bump to `1.1.0`.

## Related Issue

N/A

## Motivation and Context

This adds in a feature that Airbnb needed, and would also be useful to other developers.

## How Has This Been Tested

Simulator, XCTests

## Types of changes

## Checklist


Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/PartialMonthVisibilityDemoViewController.swift
```diff
@@ -0,0 +1,122 @@
...
+        calendarView.bottomAnchor.constraint(equalTo: view.layoutMarginsGuide.bottomAnchor),
+        calendarView.leadingAnchor.constraint(
+          greaterThanOrEqualTo: view.layoutMarginsGuide.leadingAnchor),
+        calendarView.trailingAnchor.constraint(
+          lessThanOrEqualTo: view.layoutMarginsGuide.trailingAnchor),
+        calendarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375)
+      ])
+    case .horizontal(let monthWidth):
+      NSLayoutConstraint.activate([
```
brynbodayle, line 57: Maybe we could add an alert that pops up when you open this config indicating that it does nothing since this option only works for `.vertical` layouts.

Sources/Public/MonthsLayout.swift
```diff
@@ -48,10 +72,40 @@ extension MonthsLayout: Equatable {
...
     default: return false
     }
   }
 
 }
+
+// MARK: - VerticalMonthsLayoutOptions
+
+/// Layout options for a vertically-scrolling calendar.
+public struct VerticalMonthsLayoutOptions: Equatable {
```
brynbodayle, line 86: Thoughts on using an option set here instead? Might be a little more straightforward to use in the API.

bryankeller: My only hesitation is that there might be non-boolean options that we'd want to add in the future

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/PartialMonthVisibilityDemoViewController.swift
bryankeller: Ah, I should probably just hide it when that segmented control is on horizontal. Good call!

bryankeller:
@brynbodayle thanks for the review! latest commit makes the demo VC switch between two data sources, one for horizontal one for vertical, depending on which segmented control option is selected.

Example/HorizonCalendarExample/HorizonCalendarExample/DemoPickerViewController.swift
```diff
@@ -79,22 +89,38 @@ final class DemoPickerViewController: UIViewController {
...
+    segmentedControl.addTarget(
+      self,
+      action: #selector(monthsLayoutPickerValueChanged),
+      for: .valueChanged)
     return segmentedControl
   }()
 
+  @objc
+  private func monthsLayoutPickerValueChanged() {
+    tableView.reloadData()
```
bryankeller, line 101: f-73807-Bumperhammer-213_Sharp-Sat

PR #17, opened 2020-06-27
Bk/fix scroll to item bug

bryankeller:
## Details

This fixes a bug in the `scrollToDay` and `scrollToMonth` functions. Prior to this change, scrolling to an item that was close to the content boundary could result in a temporary layout glitch. 

For example, if you scroll to the first visible month with a scroll position of `.centered`, the calendar will try to layout with the assumption that having the first visible month centered on the screen was a valid position, which it's not (the first visible month needs to be at the start of the scrollable content - it can't be centered unless you're over-scrolling / rubber banding, neither of which are valid final resting positions).

You can see an example of this bug when programmatically scrolling to the first visible day in the calendar with scroll position set to `.lastFullyVisiblePosition`

| Before | After |
| ---- | ---- |
| ezgif-6-f4210c1b82de | ezgif-6-6091550d0003 | 

2 of the commits are small refactors. If you review the commits in order, they build on one another in a pretty straightforward way.

## Related Issue

N/A

## Motivation and Context

Bug fix

## How Has This Been Tested

Tested

## Types of changes

## Checklist

### DIFF
--- a/Sources/Internal/VisibleItemsProvider.swift
+++ b/Sources/Internal/VisibleItemsProvider.swift
@@ -61,9 +61,18 @@ final class VisibleItemsProvider {
     -> LayoutItem
   {
     let baseMonthFrame = frameProvider.frameOfMonth(month, withOrigin: offset)
-    let finalMonthFrame = translatedFrame(baseMonthFrame, for: scrollPosition, offset: offset)
-    let finalFrame = frameProvider.frameOfMonthHeader(inMonthWithOrigin: finalMonthFrame.origin)
-    return LayoutItem(itemType: .monthHeader(month), frame: finalFrame)
+    let proposedMonthFrame = proposedScrollToItemFrame(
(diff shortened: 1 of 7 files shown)


Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -839,6 +849,84 @@ final class VisibleItemsProvider {
     }
   }
 
+  private func correctedScrollToItemFrameForContentBoundaries(
```
brynbodayle, line 852: unclear to me what the correction we're applying here is

bryankeller: https://github.com/airbnb/HorizonCalendar/pull/17/commits/8db3ad34e1e74dc15f841dfff6a541801736c40f added a detailed description for the purpose of this function - hoping that clears it up!

PR #20, opened 2020-07-01
Bk/fix scroll to item bug 2

bryankeller:
## Details

After I merged this https://github.com/airbnb/HorizonCalendar/pull/17, I found some other bugs and edge cases. I've improved the logic and tested extensively, including dedicated unit tests for these cases.

Ass a reminder, this is fixing the case where you tell the calendar to scroll to the first visible date with a `.centered` scroll position, for example, which is not possible since the first visible date won't be in a natural resting position if it's centered. This logic corrects for that and ensures that our target scroll position is always a valid resting state for the calendar's scroll view.

## Related Issue

N/A

## Motivation and Context

Trying to get this working perfectly!

## How Has This Been Tested

Testing on real device and added unit tests.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Internal/VisibleItemsProvider.swift
+++ b/Sources/Internal/VisibleItemsProvider.swift
@@ -872,7 +872,10 @@ final class VisibleItemsProvider {
     var currentMonthFrame = monthFrame
 
     // Look backwards for boundary-determining months
-    while bounds.contains(currentMonthFrame.origin), minimumScrollOffset == nil {
+    while
+      bounds.contains(currentMonthFrame.origin.alignedToPixels(forScreenWithScale: scale)),
+      minimumScrollOffset == nil
+    {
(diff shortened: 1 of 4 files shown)


Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -896,13 +896,14 @@ final class VisibleItemsProvider {
     currentMonth = month
     currentMonthFrame = monthFrame
     while
-      bounds.contains(CGPoint(x: currentMonthFrame.maxX, y: currentMonthFrame.maxY)),
+      bounds.contains(
+        CGPoint(x: currentMonthFrame.maxX - 1, y: currentMonthFrame.maxY - 1)
```
bryankeller, line 900: `CGRect`'s `containsPoint` function won't return `true` if the point lies on the trailing/bottom edge

Issue #18, opened 2020-06-28
Incompatible with prefersLargeTitles


davidrothera:
When added as the first subview (or only subview) to a view within a navigation controller that has the `prefersLargeTitles` property set to `true` there is an issue where the large title will show the large title but if you scroll down and back to the top the large title never comes back which is the usual behaviour.

This is reproducible using the demo app by adding the below to the `viewDidLoad` of the DemoPickerViewController:

```swift
navigationController?.navigationBar.prefersLargeTitles = true
```

bryankeller:
@davidrothera this is fixed in `v1.1.3`. Thanks again for the report 🙂

PR #21, opened 2020-07-01
Bk/scroll behavior improvements

bryankeller:
## Details

This PR fixes several issues with how the internal scroll view's state was handled. Overall, it simplifies the looping logic, and hardens the boundary-determining logic so that it's impossible to scroll fast enough that the calendar is put into a state where all visible items have been scrolled past (difficult to do, but if you scroll reallllyyy fast it's possible to skip right over a boundary-determining month).

- Simplify the `scrollMetricsMutator` (responsible for looping the content offset and setting the content inset to create boundaries). Instead of basing calculations on the current bounds / height, the `scrollMetricsMutator` now uses constants to determine its minimum offset, maximum offset, and looping region size.
- Fixes some bugs with inset handling
- Ensure that there's no way to scroll so fast that we skip right over a boundary month. This is done by looking for a boundary-determining month in the `extendedBounds`, rather than the visible bounds. The extended bounds always encapsulates the current scroll position and some item from the last frame, enabling us to be sure that we're never skipping over a large portion of content just because our scroll velocity was crazy fast.

Put all together, the changes in this PR fix https://github.com/airbnb/HorizonCalendar/issues/18 - now, things behave correctly when used with an expanding / collapsing nav bar.

| Before | After |
| ---- | ---- |
| ezgif-1-8e045c99086f | ezgif-1-544e48721a67 |

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/18

## Motivation and Context

Cleaning up and simplifying this code was overdue. This also paves the way for supporting proper automatic content inset adjustment behavior and manual content inset preferences (for which there's also an open issue).

## How Has This Been Tested

Testing on my phone and in simulator.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Internal/ScrollMetricsMutator.swift
+++ b/Sources/Internal/ScrollMetricsMutator.swift
@@ -22,15 +22,13 @@ final class ScrollMetricsMutator {
 
   // MARK: Lifecycle
 
-  init(scrollMetricsProvider: ScrollMetricsProvider, bounds: CGRect, scrollAxis: ScrollAxis) {
+  init(scrollMetricsProvider: ScrollMetricsProvider, scrollAxis: ScrollAxis) {
     self.scrollMetricsProvider = scrollMetricsProvider
-    self.bounds = bounds
     self.scrollAxis = scrollAxis
   }
 
   // MARK: Internal
 
-  let bounds: CGRect
   let scrollAxis: ScrollAxis
 
   func setUpInitialMetricsIfNeeded() {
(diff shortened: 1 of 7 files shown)


Sources/Internal/ScrollMetricsMutator.swift
```diff
@@ -108,26 +112,15 @@ final class ScrollMetricsMutator {
 
   // MARK: Private
 
-  /// A content size multiplier of 15 is large enough to prevent a high-velocity scroll from causing us to bump into an edge prematurely.
-  private static let contentSizeMultiplier: CGFloat = 15
-  private static let loopingRegionSizeDeterminingMultiplier: CGFloat = 1 / 3
+  private static let contentSize: CGFloat = 30_000 // 10,000 padding from min and max offset
+  private static let minimumContentOffset: CGFloat = 10_000 // 1/3 of content size
+  private static let maximumContentOffset: CGFloat = 20_000 // 2/3 of content size
+  private static let loopingRegionSize: CGFloat = 10_000 // Distance between min and max offset
```
bryankeller, line 118: There was no good reason for these looping key-offsets to be dynamic based on the current bounds before - it just made things way more complicated to understand and debug, since we'd have to invalidate the whole `scrollMetricsMutator` anytime our bounds changed.

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -564,8 +564,34 @@ final class VisibleItemsProvider {
...
+      let monthFrame: CGRect
+      if let cachedMonthFrame = framesForVisibleMonths[month] {
+        monthFrame = cachedMonthFrame
+      } else {
+        let monthOrigin = frameProvider.originOfMonth(containing: layoutItem)
+        monthFrame = frameProvider.frameOfMonth(month, withOrigin: monthOrigin)
+      }
+
+      // Use the calculated month frame to determine content boundaries if needed.
+      determineContentBoundariesIfNeeded(
```
bryankeller, line 580: The main change here is that we now determine our content boundaries after the `layoutItem.frame.intersects(extendedBounds)` check but not after the `layoutItem.frame.intersects(bounds)` check (as it was before this change). Determining this after the latter check was fragile since it was possible for a user to scroll so fast that they would skip right past a boundary-determining month, sending them into a blank scrolling view with no visual feedback as to what happened.

`extendedBounds` already existed to solve a similar problem, and I should have been using it for this check since the start.

Sources/Public/CalendarView.swift
```diff
@@ -127,6 +116,7 @@ public final class CalendarView: UIView {
       guard
         newValue.size != bounds.size,
         scrollToItemContext == nil,
+        !scrollView.isDragging,
```
bryankeller, line 119: This works around an issue where the calendar's bounds is changing at the same time that a user is dragging on the scroll view. This causes a weird feeling interaction, so better to guard against it.

Reminder that the point of this `bounds` `willSet` override is to do some bookkeeping so that we can maintain our scroll position after the bounds changes (like during rotation). If the user has their finger dragging on the calendar / is actively interacting with the calendar, then we don't need to do anything extra here.

Tests/VisibleItemsProviderTests.swift
```diff
@@ -557,7 +557,9 @@ final class VisibleItemsProviderTests: XCTestCase {
       "Unexpected centermost layout item.")
 
     XCTAssert(details.minimumScrollOffset == nil, "Unexpected minimum scroll offset.")
-    XCTAssert(details.maximumScrollOffset == nil, "Unexpected maximum scroll offset.")
+    XCTAssert(
```
bryankeller, line 560: This is a correct consequence of the improved boundary-determining code when a user is scrolling very fast.

PR #28, opened 2020-07-14
Make Day & Month inits public

mosaic-engineering:
## Details

Made `Day` and `Month` structs' initializers public.

## Related Issue

n/a

## Motivation and Context

Solves the case where the user wants to programmatically select a range of dates. To do so, there are two problems the user has to solve

1. Providing the right `Set<ClosedRange<Date>>` to the `DayRangeItemProvider`
2. Having the right `DayRangeItemProvider` enum to apply the appropriate selection style for start/end dates

The first problem can be easily solved by the user. However, the second one can not be solved as both `Day` and `Month` structs' initializers' levels are `internal` and are not accessible to the user.

## How Has This Been Tested

No significant changes have been made to the codebase. Therefore believe that no additional tests have to be written.

## Types of change

## Checklist

### DIFF
--- a/Sources/Public/Day.swift
+++ b/Sources/Public/Day.swift
@@ -23,7 +23,7 @@ public struct Day {
 
   // MARK: Lifecycle
 
-  init(month: Month, day: Int) {
+  public init(month: Month, day: Int) {
     self.month = month
     self.day = day
   }
--- a/Sources/Public/Month.swift
+++ b/Sources/Public/Month.swift
@@ -23,7 +23,7 @@ public struct Month {
 
   // MARK: Lifecycle
 
-  init(era: Int, year: Int, month: Int, isInGregorianCalendar: Bool) {
+  public init(era: Int, year: Int, month: Int, isInGregorianCalendar: Bool) {
     self.era = era
     self.year = year
     self.month = month


bryankeller:
Hi @mosaic-engineering , thanks for the proposed change. I think this is something that I could probably make clearer in the documentation. With that said...

These two structs are intentionally internal to the calendar, as only `HorizonCalendar` has the necessary pieces of information (a `Date` range and a `Calendar` instance) to be able to construct these. `HorizonCalendar` ensures that only valid `Day` / `Month` structs are created by always deriving them from the `Calendar` instance that you initialized `CalendarViewContent` with. If any consumer could construct these, then there could be programmer errors where invalid `Day`s are created. For example, `Day(year: 2020, month: 01, day: 500)` makes no sense.

The basic philosophy is that consumers of the API always provide `Date`s and a `Calendar` instance to `HorizonCalendar`, and `HorizonCalendar` always passes `Day` and `Month` instances to you that are guaranteed to be valid for the current date range and calendar combination.

----

> Having the right DayRangeItemProvider enum to apply the appropriate selection style for start/end dates

I think the key to solving your problem is knowing that you can always convert a `Day` to a `Date`. You can use `let date = calendar.date(from: day.dateComponents)!` to convert a selected day into a corresponding date. Make sure to use the same `calendar` instance that you used to initialize the `CalendarViewContent` otherwise you'll have some weird bugs due to calendar conversion issues!

I also think the example project code could be useful. I do a conversion from `Day` -> `Date` here for the purpose of using the `withDayRangeItemProvider` https://github.com/airbnb/HorizonCalendar/blob/master/Example/HorizonCalendarExample/HorizonCalendarExample/Demo%20View%20Controllers/DayRangeSelectionDemoViewController.swift#L121

----

Let me know if that solves your issue 🙂

mosaic-engineering:
>  If any consumer could construct these, then there could be programmer errors where invalid Days are created.

Makes sense. Framework enforcing rules to prevent programmer errors. Brilliant!

Thanks for the feedback and keep up the great works guys :)

Issue #19, opened 2020-06-30
 How can I add bottom contentInsets scroll view?


misha8414:
Hello, is it possible to add bottom contentInsets for scroll view?

bryankeller:
Hi @misha8414 - I'm working on supporting this soon. Until then, a temporary workaround might be to constrain your calendar to the layout margins so that the content isn't overlapped.

JulienLevallois:
I have the same problem. I need top & bottom contentInsets for the scrollview

alex-sarkisov:
For now I added top inset, setting the month bottom inset to 0
> .withMonthDayInsets(UIEdgeInsets(top: 20, left: 20, bottom: 0, right: 20))
and adding top padding to the custom month view title
> titleLabel.topAnchor.constraint(equalTo: topAnchor, constant: 20)

bryankeller:
@misha8414 @JulienLevallois @alex-sarkisov I've added the ability to set insets using `UIView`'s `layoutMargins` and `directionalLayoutMargins` properties. The changes are included in `v1.2.0`. Thank you again for reporting this and for your patience as I investigated a good way to do this 🙂

PR #35, opened 2020-07-28
Bk/inset support

bryankeller:
## Details

This adds support for insetting the content without affecting the scrollable / user-interact-able region using the `layoutMargins` and `directionalLayoutMargins` properties on `CalendarView.swift`

The approach of using layout margins seems to work pretty well - one nice thing we get for free is automatic safe area inset respecting (which can be turned off via the flag on `UIView`). The one quirk is that we don't support negative layout margins, since this causes really weird scrolling / layout logic. I can't really think of a valid reason to use negative values either, but hopefully I'm not missing something obvious.

One other caveat - For horizontal layout calendars, the current sizing logic doesn't take into account the vertical dimension of the frame. Instead, it's determined by the `monthWidth` that's passed into `CalendarViewContent`. In hindsight, this probably isn't how the horizontal layout should work - instead, the horizontal layout should determine its month width by looking at the available height. Because of this shortcoming with how the horizontal layout works, there's not a great way to make layout margins work perfectly in the vertical dimension, since the available space in the vertical dimension isn't taken into account. I don't think this is a showstopper, as the leading / trailing / top margins all work well. It's only when you use both a top and bottom margin that the latter is basically just ignored. As part of the 2.0 release of HorizonCalendar, I'll change this API / behavior so that it works in a more predictable way, which will have the consequence of making the layout margins work correctly for top / bottom when using a horizontal layout.

| Vertical | Horizontal (just leading / trailing margins) |
| ---- | ---- |
| ezgif-6-41b58e78f50a | ezgif-6-eaf73940399e |

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/19

## Motivation and Context

It's common for people to want to inset their calendar's content without adjusting the frame or scrollable region. For example, this could be useful when displaying a vertically scrolling calendar on iPad, where the content would look way too big if it went edge-to-edge horizontally. By setting layout margins, the content of the calendar can be inset while still allowing scroll gestures all the way to the edge of the screen.

## How Has This Been Tested

Tested on simulator and device.

## Types of changes

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ A declarative, performant, calendar UI component that supports use cases ranging
 Features:
 
 - Supports all calendars from `Foundation.Calendar` (Gregorian, Japanese, Hebrew, etc.)
-- Displaying months in a vertically-scrolling or horizontally-scrolling layout
+- Display months in a vertically-scrolling or horizontally-scrolling layout
 - Declarative API that enables unidirectional data flow for updating the content of the calendar
 - A custom layout system that enables virtually infinite date ranges without increasing memory usage
 - Specify custom views for individual days, month headers, and days of the week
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -23,9 +23,10 @@ Features:
 - A day selection handler to monitor when a day is tapped
 - Customizable layout metrics
 - Pinning days of the week to the top
-- Showing partial boundary months (exactly 2020-03-14 to 2020-04-20, for example)
-- Scrolling to arbitrary dates and months, with or without animation
+- Show partial boundary months (exactly 2020-03-14 to 2020-04-20, for example)
```
brynbodayle, line 26: grammar++

Sources/Public/CalendarView.swift
```diff
@@ -150,6 +172,10 @@ public final class CalendarView: UIView {
     }
   }
 
+  public override func layoutMarginsDidChange() {
+    setNeedsLayout()
```
brynbodayle, line 176: Feels like good practice to call super even if the docs say it doesn't do anything.

PR #39, opened 2020-08-04
Bk/fix horizontal layout clipping

bryankeller:
## Details

I noticed an issue where month headers were getting clipped too soon when figuring out which items are visible / not visible in a horizontal layout calendar. The root issue is that my heuristic for figuring out what intersects the visible rect was wrong for horizontal layouts.

| Before | After |
| ---- | ---- |
| ezgif-4-ed9a537cf4de | ezgif-4-e3ec00150639 |

## Related Issue

N/A

## Motivation and Context

Bug fix.

## How Has This Been Tested

iOS sim, unit test.

## Types of changes

## Checklist

- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/Sources/Internal/VisibleItemsProvider.swift
+++ b/Sources/Internal/VisibleItemsProvider.swift
@@ -142,8 +142,6 @@ final class VisibleItemsProvider {
     let maxY = max(bounds.maxY, previouslyVisibleLayoutItem.frame.maxY)
     let extendedBounds = CGRect(x: minX, y: minY, width: maxX - minX, height: maxY - minY)
 
-    var numberOfConsecutiveNonIntersectingItems = 0
-
     var handledDayRanges = Set<DayRange>()
 
     var originsForMonths = [Month: CGPoint]()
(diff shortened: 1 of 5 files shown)


Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -585,24 +573,25 @@ final class VisibleItemsProvider {
     handledDayRanges: inout Set<DayRange>,
     shouldStop: inout Bool)
   {
-    if layoutItem.frame.intersects(extendedBounds) {
-      numberOfConsecutiveNonIntersectingItems = 0
+    let month = layoutItem.itemType.month
 
-      let month = layoutItem.itemType.month
+    // Calculate and the current month frame if it's not cached; it will be used in other
```
brynbodayle, line 578: typo here?

Sources/Public/MonthsLayout.swift
```diff
@@ -33,6 +33,13 @@ public enum MonthsLayout {
 
   // MARK: Internal
 
+  var isHorizontal: Bool {
```
brynbodayle, line 36: Feels like this kind of goes against the purpose of an enum/switch, but I guess it's just to avoid `case let` syntax.

bryankeller: Yeah, it's only exposed internally but yes, it's a way to avoid an awkward case let in the `VisibleItemsProvider`.

PR #41, opened 2020-08-04
Bk/fix scroll to item crash

bryankeller:
## Details

This fixes a crash than can occur if the content of the calendar is set while it's programmatically scrolling. In this scenario, the display link does not get invalidated even though our `scrollToItemContext` is set to `nil`. If the `scrollToItemContext` is `nil` the next time the display link fires, then we hit a precondition failure.

To simplify things, we should really just invalidate the display link whenever our `scrollToItemContext` is set. This PR makes that change.

## Related Issue

N/A

## Motivation and Context

Crash fix.

## How Has This Been Tested
iOS simulator example app and Airbnb app (where the  repro case was identified).

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -168,7 +168,7 @@ public final class CalendarView: UIView {
     super.didMoveToWindow()
 
     if window == nil {
-      scrollToItemDisplayLink?.invalidate()
+      scrollToItemContext = nil
     }
   }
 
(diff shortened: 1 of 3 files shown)


Sources/Public/CalendarView.swift
```diff
@@ -688,10 +691,6 @@ extension CalendarView: UIScrollViewDelegate {
 
     let isUserInitiatedScrolling = scrollView.isDragging && scrollView.isTracking
 
-    if isUserInitiatedScrolling, scrollToItemDisplayLink != nil {
-      scrollToItemDisplayLink?.invalidate()
```
bryankeller, line 692: `scrollToItemContext` is set to `nil` just a few lines down, so no need to manually invalidate here anymore.

Issue #31, opened 2020-07-19
Is it possible to add a "footer" to the bottom of each month on the vertical calendar?


seanperez29:
Similar to the ability to add a CalendarItem to customize the month header, is there the capability to add a footer at the bottom of each month on the vertical scrolling calendar? (Example in screenshot). Or will this possibly be added if that capability doesn't currently exist? Thanks! Awesome library btw!
<img width="372" alt="Screen Shot 2020-07-18 at 8 01 22 PM" src="https://user-images.githubusercontent.com/18096448/87865352-9cd4e580-c931-11ea-8f5a-6e8b9e86e202.png">

bryankeller:
Hey @seanperez29 ! This is a cool suggestion, I could see us using something like this in the Airbnb app one day.

This isn't something that's currently on the roadmap, but I think this would be a great API addition if it's something you'd be interested in tackling. I've tagged this as being a good issue for first-time contributors!

----

If anyone wants to look into implementing this, it would probably be a new item provider on `CalendarViewContent`  (some kind of footer item type).

The more complicated parts are deciding how to lay something like this out. Does it have its own padding values? Does it self-size, or statically size? Is this a one-off item provider type, or part of a more general system for adding supplementary content to various parts of the calendar - a month footer being just one of several supported locations.

If we can think through a great API that will serve us well for this feature / potentially others in the future, that would be a great first step!

mustiikhalil:
@bryankeller is this issue still open? I would love to help out and add the footer for the view. If so, let's brainstorm a way to make it happen

bryankeller:
@mustiikhalil it is still open, and any help would be greatly appreciated since there are some other things I'm working on unrelated to this currently (proper RTL support being top priority for me).

The way I implemented headers was by forcing them to be self-sized. This was great for Airbnb's use case (and frankly, forces people to support dynamic / accessible type 😛), but perhaps it would be valuable to generalize the sizing of headers and footers so that people can choose their own sizing preference.

The general approach here is to copy how headers are handled at both an API and implementation level, with one caveat - we should add the ability to specify a sizing mode. My initially feeling is that the sizing mode should be specified on `CalendarItemModel` that way you can choose to self-size or statically-size headers and footers on a case-by-case basis, but `CalendarItemModel` is also used for for items, which are sized by the calendar and therefore having a sizing mode makes no sense for those.

[...]

PR #75, opened 2020-11-20
feat: provide ability to add footer to month

saudeon:
## Details

- Adds in the ability to provide a footer for a month.
- Provides a default footer with an empty view.

## Related Issue

#31

## Motivation and Context

A request for being able to provide views for month footers was made.

## How Has This Been Tested

- Made sure that original tests compile and run.

## Types of changes

## Checklist


saudeon:
@bryankeller I've noticed that the first footer always shows, but I am required to scroll past a certain point before the next footer is shown. Did I miss something extremely obvious?

bryankeller:
Hey @saudeon - thanks a ton for working on this. I'm currently taking some time off but really excited about this change. I'll get back to you in a couple days when I have a chance to go through the code. Hopefully the bug you're describing is an easy fix 🙂 thanks again got tackling this!

bryankeller:
> I've noticed that the first footer always shows, but I am required to scroll past a certain point before the next footer is shown. Did I miss something extremely obvious?

Perhaps no obvious but I see the problem! Because the layout is built incrementally, looking both forwards and backwards from the location of a known item, we need to be able to describe the ordering of items (including footers) both looking forwards and backwards from any given item. More details here https://github.com/airbnb/HorizonCalendar/blob/master/Docs/TECHNICAL_DETAILS.md#architecture-walkthrough

The thing that's missing is in `private func nextItemType(from itemType: LayoutItem.ItemType) -> LayoutItem.ItemType`. That implementation of that function currently says "the next item after the last day of the month is the next month's header" - when really it should say "the next item after the last day of the month is either the same month's footer (or the next month's header if there is no footer)" ... (more on that last part in my next message).

bryankeller:
The other high level piece of feedback is that the month footer should actually be optional (rather than always present but given 0 height). I think we can probably have the `LayoutItemTypeEnumerator` always say:

- Month footers come before the month header of a subsequent month when looking backwards
- Month footers come after the last day of the previous month when looking forwards

However, in the `VisibleItemsProvider`, we should *not* create a `VisibleCalendarItem` / do any work if there's no month footer item provider set on `CalendarViewContent`.

I've left some individual comments on the PR which assume we make the month footer item provider optional.

----


Thanks again for your hard work on this!

Sources/Internal/DefaultItemProviders.swift
```diff
@@ -142,5 +142,33 @@ extension CalendarViewContent {
         accessibilityTraits: []),
       viewModel: .init(text: dayText, accessibilityLabel: accessibilityLabel))
   }
-
+  
+  static func defaultMonthFooterItemModelProvider(
```
bryankeller, line 146: if we make the month footer optional, there's no reason to provide a default item. I also think most people won't need this, or if they do, they'll want something very custom, so providing a default doesn't provide a ton of value.

saudeon: Already en route to removing this default provider

Sources/Internal/FrameProvider.swift
```diff
@@ -27,13 +27,15 @@ final class FrameProvider {
     size: CGSize,
     layoutMargins: NSDirectionalEdgeInsets,
     scale: CGFloat,
-    monthHeaderHeight: CGFloat)
+    monthHeaderHeight: CGFloat,
+    monthFooterHeight: CGFloat = 0.0)
```
bryankeller, line 31: this can be `CGFloat?`. We can probably leave out the default value since we always initialize this with all values anyways.

Sources/Internal/FrameProvider.swift
```diff
@@ -247,6 +268,7 @@ final class FrameProvider {
   private let content: CalendarViewContent
   private let monthWidth: CGFloat
   private let monthHeaderHeight: CGFloat
+  private let monthFooterHeight: CGFloat
```
bryankeller, line 271: same change: `CGFloat?`

Sources/Internal/FrameProvider.swift
```diff
@@ -152,6 +163,16 @@ final class FrameProvider {
     return CGRect(origin: CGPoint(x: x, y: y), size: daySize)
   }
 
+  func frameOfMonthFooter(_ month: Month, inMonthWithOrigin monthOrigin: CGPoint) -> CGRect {
```
bryankeller, line 166: we can omit the `month` parameter here. I don't believe it's used.

saudeon: Month is used here to determine how many rows for the month are being displayed

Sources/Internal/FrameProvider.swift
```diff
@@ -273,6 +295,7 @@ final class FrameProvider {
         daySize.height - content.verticalDayMargin -
         (maxNumberOfWeekRowsPerMonth * daySize.height) -
         ((maxNumberOfWeekRowsPerMonth - 1) * content.verticalDayMargin) -
+        monthFooterHeight -
         content.monthDayInsets.bottom
```
bryankeller, line 299: Can we swap the order of this subtraction so that it reflects the actual ordering of the values? Basically, the `monthDayInsets.bottom` should be between the days and the month footer, not below the month footer. Like this:

image

Sources/Internal/FrameProvider.swift
```diff
@@ -79,6 +81,15 @@ final class FrameProvider {
...
       let y = minYOfMonth(containingDayItemWithFrame: layoutItem.frame, atRowInMonth: rowInMonth)
       return CGPoint(x: x, y: y)
+        
+    case .monthFooter(let month):
+      let x = layoutItem.frame.origin.x
+        let y = layoutItem.frame.minY -
+            monthHeaderHeight -
+            content.monthDayInsets.top -
+            (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) -
+            (CGFloat(numberOfRows(in: month)) * (daySize.height + content.verticalDayMargin))
```
bryankeller, line 91: Similar feedback: this will need to account for `monthDayInsets.bottom`, since that pushes the footer down by some amount. See https://github.com/airbnb/HorizonCalendar/pull/75/files#r531198063

bryankeller: Also, I think

`- (numberOfRows * (daySize.height + content.verticalDayMargin))`

needs to be

`- ((numberOfRows * daySize.height) + ((numberOfRows - 1) * content.verticalDayMargin))`

because if there are 5 rows of days in a month, there are only 4 day margins. As it's written now, there would be some extra margin added. I made this mistake for a while and fixed it a few versions ago - it's a subtle one 🙂

Sources/Internal/FrameProvider.swift
```diff
@@ -152,6 +163,16 @@ final class FrameProvider {
...
   }
 
+  func frameOfMonthFooter(_ month: Month, inMonthWithOrigin monthOrigin: CGPoint) -> CGRect {
+    let x = monthOrigin.x
+    let y = monthOrigin.y +
+        monthHeaderHeight +
+        content.monthDayInsets.top +
+        (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) +
+        (CGFloat(numberOfRows(in: month)) * (daySize.height + content.verticalDayMargin))
+    return CGRect(origin: CGPoint(x: x, y: y), size: CGSize(width: monthWidth, height: monthFooterHeight))
```
bryankeller: This `y` calculation should include the `monthDayInsets.bottom`, for the same reason explained here https://github.com/airbnb/HorizonCalendar/pull/75/files#r531198063

Sources/Internal/FrameProvider.swift
```diff
@@ -323,6 +346,7 @@ final class FrameProvider {
       (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) +
       (CGFloat(numberOfRows) * daySize.height) +
       (CGFloat(numberOfRows - 1) * content.verticalDayMargin) +
+      monthFooterHeight +
```
bryankeller: same as https://github.com/airbnb/HorizonCalendar/pull/75/files#r531198063

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -29,6 +29,7 @@ final class VisibleItemsProvider {
     layoutMargins: NSDirectionalEdgeInsets,
     scale: CGFloat,
     monthHeaderHeight: CGFloat,
+    monthFooterHeight: CGFloat,
```
bryankeller, line 32: make this optional as well

Sources/Public/CalendarViewContent.swift
```diff
@@ -321,6 +352,7 @@ public final class CalendarViewContent {
 
   // TODO(BK): Make all item provider closures private(set) after legacy `CalendarItem` is removed.
   var monthHeaderItemModelProvider: (Month) -> InternalAnyCalendarItemModel
+  var monthFooterItemModelProvider: (Month) -> InternalAnyCalendarItemModel
```
bryankeller: make this an optional closure and remove the default value

Sources/Internal/FrameProvider.swift
bryankeller: Ah! I missed that every time I looked, whoops!

Sources/Internal/DefaultItemProviders.swift
saudeon: Removed

Sources/Internal/FrameProvider.swift
saudeon: Updated to optional

Sources/Internal/FrameProvider.swift
saudeon: Updated to optional

Sources/Internal/FrameProvider.swift
saudeon: Updated as per comments

Sources/Internal/FrameProvider.swift
saudeon: Updated to subtle change. Subtle change indeed, thanks!

Sources/Internal/FrameProvider.swift
saudeon: Updated to include `monthDayInsets.bottom`

Sources/Internal/FrameProvider.swift
saudeon: Updated to include `monthDayInsets.bottom`

Sources/Internal/VisibleItemsProvider.swift
saudeon: Updated to optional

Sources/Public/CalendarViewContent.swift
saudeon: Updated to optional and removed default value, now is set to nil

saudeon:
Have gone through and updated majority of code as per comments, I also added a demo view controller to showcase providing a `FooterView`.

I did notice that there is sometimes some jitteriness when it comes to the scrolling now... But everything is shown as expected if it is provided!

bryankeller:
Thanks @saudeon - I can investigate the jitter. Will get back to you kn a few days. Thanks again for your hard work on this, it looks great 🙂

saudeon:
Thanks for the help getting me on the right path! Like I've mentioned, thoroughly impressed by this and as I am already using it in a few of my projects, I am more than happy to spend time contributing! 

Hope your Thanksgiving weekend was enjoyable!

Sources/Internal/FrameProvider.swift
```diff
@@ -150,6 +162,16 @@ final class FrameProvider {
...
   }
 
+  func frameOfMonthFooter(_ month: Month, inMonthWithOrigin monthOrigin: CGPoint) -> CGRect {
+    let x = monthOrigin.x
+    let y = monthOrigin.y +
+        monthHeaderHeight +
+        content.monthDayInsets.top +
+        (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) +
+        (CGFloat(numberOfRows(in: month)) * (daySize.height + content.verticalDayMargin))
+    return CGRect(origin: CGPoint(x: x, y: y), size: CGSize(width: monthWidth, height: monthFooterHeight ?? 0))
```
bryankeller, line 172:
```suggestion
    let x = monthOrigin.x
    let y = monthOrigin.y +
      monthHeaderHeight +
      content.monthDayInsets.top +
      (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) +
      ((CGFloat(numberOfRows(in: month)) * daySize.height) + (CGFloat(numberOfRows(in: month) - 1) * content.verticalDayMargin)) +
      content.monthDayInsets.bottom
    return CGRect(x: x, y: y, width: monthWidth, height: monthFooterHeight ?? 0)
```

This will fix the jitter - this calculation wasn't properly taking into account the height of number of week rows and was also missing the `content.monthDayInsets.bottom`. Easy fix 🙂

Sources/Internal/FrameProvider.swift
```diff
@@ -77,6 +79,16 @@ final class FrameProvider {
       let x = minXOfMonth(containingItemWithFrame: layoutItem.frame, at: position)
       let y = minYOfMonth(containingDayItemWithFrame: layoutItem.frame, atRowInMonth: rowInMonth)
       return CGPoint(x: x, y: y)
+        
+    case .monthFooter(let month):
+      let x = layoutItem.frame.origin.x
+        let y = layoutItem.frame.minY -
```
bryankeller, line 85: The indentation is a bit off here. I need to update the style guidelines so that contributors use 2-space indentation instead of the default 4. Personally, I like 4, but all of Airbnb's iOS stuff is 2, so I'd like to stick with that for now.

saudeon: I personally also like 4, but can fix the indention now!

Sources/Internal/FrameProvider.swift
saudeon: Ah great, I knew I missed something simple.

I wonder if there is a way to make this less intense, the calculations etc...

bryankeller: Also, these complicated calculation are simplified a bit after I merge https://github.com/airbnb/HorizonCalendar/pull/79. Once you pull that into your branch, we can update this to be something like this:

```swift
    let numberOfWeekRows = self.numberOfWeekRows(in: month)

    let x = monthOrigin.x
    let y = monthOrigin.y +
      monthHeaderHeight +
      content.monthDayInsets.top +
      heightOfDaysOfTheWeekRowInMonth() +
      heightOfDayContent(forNumberOfWeekRows: numberOfWeekRows) +
      content.monthDayInsets.bottom
    return CGRect(x: x, y: y, width: monthWidth, height: monthFooterHeight ?? 0)
```

Sources/Internal/FrameProvider.swift
```diff
@@ -77,6 +79,16 @@ final class FrameProvider {
...
+            monthHeaderHeight -
+            content.monthDayInsets.top -
+            (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) -
+            CGFloat((CGFloat(numberOfRows(in: month)) * daySize.height) + (CGFloat(numberOfRows(in: month) - 1) * content.verticalDayMargin)) -
+            content.monthDayInsets.bottom
```
bryankeller, line 90: Similar feedback as here https://github.com/airbnb/HorizonCalendar/pull/75/files#r533051542

bryankeller: Be careful of the sign though, this one is subtracting parts of the layout, while the linked comment is adding parts of the layout. It'll be pretty obvious if it's wrong though when you build and run ;)

Sources/Internal/FrameProvider.swift
saudeon: That's what I am talking about! Very nice indeed!

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SingleDaySelectionFooterDemoViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import HorizonCalendar
+
+class SingleDaySelectionFooterDemoViewController: DemoViewController {
```
bryankeller, line 19:
```suggestion
final class MonthFootersDemoViewController: DemoViewController {
```

I'd prefer to keep the demos as focused as possible, rather than combining independent concepts like single day selection and month footer configuration. Could we make the following changes (in addition to the naming suggestion above):

1. Remove all selection related naming / code
2. Give the footer some more interesting content, rather than just repeating the date in YYYY-MM format. Curious if you have any ideas / inspiration from your own projects of what would be useful in a footer?
3. Increase the value of `monthDayInsets.bottom` so that the footer isn't right up against the last week row of the month
4. Update the indentation to be 2 spaces instead of 4

Example/HorizonCalendarExample/HorizonCalendarExample/FooterView.swift
```diff
@@ -0,0 +1,48 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import HorizonCalendar
+import UIKit
+
+// MARK: CalendarItemViewRepresentable
+
+class FooterView: CalendarItemViewRepresentable {
```
bryankeller, line 21: nit:
```suggestion
final class FooterView: CalendarItemViewRepresentable {
```

Example/HorizonCalendarExample/HorizonCalendarExample/FooterView.swift
```diff
@@ -0,0 +1,48 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import HorizonCalendar
+import UIKit
+
+// MARK: CalendarItemViewRepresentable
```
bryankeller, line 19:
```suggestion
// MARK: - FooterView
```

bryankeller: And for consistency, add the `CalendarItemViewRepresentable` conformance below so that it looks similar to the `DayView` file structure.

Sources/Internal/DefaultItemProviders.swift
```diff
@@ -142,5 +142,5 @@ extension CalendarViewContent {
         accessibilityTraits: []),
       viewModel: .init(text: dayText, accessibilityLabel: accessibilityLabel))
   }
-
+
```
bryankeller, line 145: nit: remove this diff / added space

Sources/Internal/FrameProvider.swift
```diff
@@ -27,13 +27,15 @@ final class FrameProvider {
     size: CGSize,
     layoutMargins: NSDirectionalEdgeInsets,
     scale: CGFloat,
-    monthHeaderHeight: CGFloat)
+    monthHeaderHeight: CGFloat,
+    monthFooterHeight: CGFloat? = 0.0)
```
bryankeller, line 31:
```suggestion
    monthFooterHeight: CGFloat?)
```

Sources/Internal/LayoutItemTypeEnumerator.swift
```diff
@@ -102,6 +103,10 @@ final class LayoutItemTypeEnumerator {
       } else {
         return .day(calendar.day(byAddingDays: -1, to: day))
       }
+    
+    case .monthFooter(let month):
+        let day = calendar.day(containing: calendar.lastDate(of: month))
+        return .day(day)
```
bryankeller, line 109: Ah, some more 4 space instead of 2 space indentation. I'll look into enforcing this at the project level. This is probably annoying to deal with, sorry!

saudeon: It's okay! As for enforcing, can look into `swiftlint` or something similar?

bryankeller: yeah, great idea. We already use that for the Airbnb app repo internally, and it pulls rules from our Swift style guide: https://github.com/airbnb/swift

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SingleDaySelectionFooterDemoViewController.swift
saudeon: 1. ✅ 
2. I haven't found the need just yet to add a footer to these in my own projects, but I can definitely make an attempt at coming up with something...
3. ✅ 
4. ✅

Sources/Internal/FrameProvider.swift
saudeon: Once #79 is merged, I'll update this branch with the same fixes to accommodate the clean up.

Sources/Internal/FrameProvider.swift
saudeon: This is all working as expected now.
I will spend some time on implementing some form of footer experience that I could see worthy of this.
Do you want me to solely focus on UI presentation of a footer, or actually the UI and logic implementation?

Can see it getting quite intense just for a demo that is all, I'm indifferent, just want some direction on how thought out I should be in the demo.

Sources/Internal/FrameProvider.swift
saudeon: `master` has been merged into this branch, with the changes to accommodate the cleaner frame math

Sources/Internal/FrameProvider.swift
```diff
@@ -77,6 +79,16 @@ final class FrameProvider {
...
+        
+    case .monthFooter(let month):
+      let x = layoutItem.frame.origin.x
+      let y = layoutItem.frame.minY -
+        monthHeaderHeight -
+        content.monthDayInsets.top -
+        (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) -
+        CGFloat((CGFloat(numberOfRows(in: month)) * daySize.height) + (CGFloat(numberOfRows(in: month) - 1) * content.verticalDayMargin)) -
+        content.monthDayInsets.bottom
+      return CGPoint(x: x, y: y)
```
bryankeller, line 91:
```suggestion
    case .monthFooter(let month):
      let numberOfWeekRows = self.numberOfWeekRows(in: month)

      let x = layoutItem.frame.origin.x
      let y = layoutItem.frame.minY -
        content.monthDayInsets.bottom -
        heightOfDayContent(forNumberOfWeekRows: numberOfWeekRows) -
        heightOfDaysOfTheWeekRowInMonth() -
        content.monthDayInsets.top -
        monthHeaderHeight
      return CGPoint(x: x, y: y)
```

Adopting new helper functions and swapping the order of subtraction so it reads like we'd subtracting each adjacent piece upwards until we get to our month's min Y value.

Sources/Internal/FrameProvider.swift
```diff
@@ -152,6 +164,17 @@ final class FrameProvider {
...
+  func frameOfMonthFooter(_ month: Month, inMonthWithOrigin monthOrigin: CGPoint) -> CGRect {
+    let x = monthOrigin.x
+    let y = monthOrigin.y +
+      monthHeaderHeight +
+      content.monthDayInsets.top +
+      (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) +
+      ((CGFloat(numberOfRows(in: month)) * daySize.height) + (CGFloat(numberOfRows(in: month) - 1) * content.verticalDayMargin)) +
+      content.monthDayInsets.bottom
+      return CGRect(x: x, y: y, width: monthWidth, height: monthFooterHeight ?? 0)
+  }
```
bryankeller, line 176:
```suggestion
  func frameOfMonthFooter(_ month: Month, inMonthWithOrigin monthOrigin: CGPoint) -> CGRect {
    let numberOfWeekRows = self.numberOfWeekRows(in: month)

    let x = monthOrigin.x
    let y = monthOrigin.y +
      monthHeaderHeight +
      content.monthDayInsets.top +
      heightOfDaysOfTheWeekRowInMonth() +
      heightOfDayContent(forNumberOfWeekRows: numberOfWeekRows) +
      content.monthDayInsets.bottom
      return CGRect(x: x, y: y, width: monthWidth, height: monthFooterHeight ?? 0)
  }
```

Adopting new helpers

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+import HorizonCalendar
+
+class MonthFootersDemoViewController: DemoViewController {
```
bryankeller, line 19:
```suggestion
final class MonthFootersDemoViewController: DemoViewController {
```

bryankeller: (small nit, I tend to use `final` for everything unless we're trying to write something subclass-able.

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+    super.viewDidLoad()
+    
+    title = "Month Footer Demo"
+    
+    calendarView.daySelectionHandler = { [weak self] day in
+      guard let self = self else { return }
+      
+      self.selectedDay = day
+      self.calendarView.setContent(self.makeContent())
+    }
```
bryankeller, line 33:
```suggestion
```

To keep things really simple, let's remove the selection logic from this file since it just distracts from the month footer stuff for someone trying to figure out how to use that.

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+  }
+  
+  override func makeContent() -> CalendarViewContent {
+    let startDate = calendar.date(from: DateComponents(year: 2020, month: 01, day: 01))!
+    let endDate = calendar.date(from: DateComponents(year: 2021, month: 12, day: 31))!
+    
+    let selectedDay = self.selectedDay
+
```
bryankeller, line 41:
```suggestion
```

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+        if let date = self?.calendar.date(from: day.components) {
+          dayAccessibilityText = self?.dayDateFormatter.string(from: date)
+        } else {
+          dayAccessibilityText = nil
+        }
+        
+        return CalendarItemModel<DayView>(
+          invariantViewProperties: .init(textColor: textColor, isSelectedStyle: day == selectedDay),
+          viewModel: .init(dayText: "\(day.day)", dayAccessibilityText: dayAccessibilityText))
+      }
```
bryankeller, line 70:
```suggestion
```

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+          textColor = .label
+        } else {
+          textColor = .black
+        }
+        
+        return CalendarItemModel<FooterView>(
+          invariantViewProperties: .init(textColor: textColor),
+          viewModel: .init(monthText: "\(month.description)")
+        )
+
```
bryankeller, line 84:
```suggestion
        return CalendarItemModel<FooterView>(
          invariantViewProperties: .init(textColor: textColor),
          viewModel: .init(monthText: "\(month.description)"))
```

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+          viewModel: .init(monthText: "\(month.description)")
+        )
+        
+      }
+  }
+  
+  // MARK: Private
+  
+  private var selectedDay: Day?
+
```
bryankeller, line 91:
```suggestion
```

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
```diff
@@ -0,0 +1,93 @@
...
+import HorizonCalendar
+
+class MonthFootersDemoViewController: DemoViewController {
+  
+  // MARK: Internal
+  
+  override func viewDidLoad() {
+    super.viewDidLoad()
+    
+    title = "Month Footer Demo"
```
bryankeller, line 26:
```suggestion
    title = "Month Footers"
```

Example/HorizonCalendarExample/HorizonCalendarExample/DemoPickerViewController.swift
```diff
@@ -68,6 +68,7 @@ final class DemoPickerViewController: UIViewController {
     ("Large Day Range", LargeDayRangeDemoViewController.self),
     ("Scroll to Day With Animation", ScrollToDayWithAnimationDemoViewController.self),
     ("Partial Month Visibility", PartialMonthVisibilityDemoViewController.self),
+    ("Month Footer Demo", MonthFootersDemoViewController.self)
```
bryankeller, line 71:
```suggestion
    ("Month Footers", MonthFootersDemoViewController.self)
```

Example/HorizonCalendarExample/HorizonCalendarExample/DemoPickerViewController.swift
```diff
@@ -76,6 +77,7 @@ final class DemoPickerViewController: UIViewController {
     ("Selected Day Tooltip", SelectedDayTooltipDemoViewController.self),
     ("Large Day Range", LargeDayRangeDemoViewController.self),
     ("Scroll to Day With Animation", ScrollToDayWithAnimationDemoViewController.self),
+    ("Month Footer Demo", MonthFootersDemoViewController.self)
```
bryankeller, line 80:
```suggestion
    ("Month Footers", MonthFootersDemoViewController.self)
```

Sources/Internal/LayoutItem.swift
```diff
@@ -57,6 +59,8 @@ extension LayoutItem.ItemType: Comparable {
       return lhsMonth <= rhsMonth
     case let (.monthHeader(lhsMonth), .day(rhsDay)):
       return lhsMonth <= rhsDay.month
+    case let (.monthHeader(lhsMonth), .monthFooter(rhsMonth)):
+      return lhsMonth < rhsMonth
```
bryankeller, line 63:
```suggestion
      return lhsMonth <= rhsMonth
```

A month header's position "is less than" / "comes before" a month footer's position even for a month header and footer in the same month.

As this is written now, `.monthHeader(2020-06) < .monthFooter(2020-06) == false`, which should not be the wrong. Let me know if this makes sense!

Sources/Internal/LayoutItem.swift
```diff
@@ -66,13 +70,26 @@ extension LayoutItem.ItemType: Comparable {
       return lhsMonth < rhsMonth
     case let (.dayOfWeekInMonth(_, lhsMonth), .day(rhsDay)):
       return lhsMonth <= rhsDay.month
+    case let (.dayOfWeekInMonth(_, lhsMonth), .monthFooter(rhsMonth)):
+      return lhsMonth < rhsMonth
```
bryankeller, line 74:
```suggestion
      return lhsMonth <= rhsMonth
```

Same thing here; The day of the week items are at the beginning of the month, so even for the same month, these come "before" the footer, which is the last item.

Sources/Internal/LayoutItem.swift
```diff
@@ -66,13 +70,26 @@ extension LayoutItem.ItemType: Comparable {
...
 
     case let (.day(lhsDay), .day(rhsDay)):
       return lhsDay < rhsDay
     case let (.day(lhsDay), .monthHeader(rhsMonth)):
       return lhsDay.month < rhsMonth
     case let (.day(lhsDay), .dayOfWeekInMonth(_, rhsMonth)):
       return lhsDay.month < rhsMonth
+    case let (.day(lhsDay), .monthFooter(rhsMonth)):
+      return lhsDay.month < rhsMonth
```
bryankeller, line 83:
```suggestion
      return lhsDay.month <= rhsMonth
```

Same reasoning here! Even in the same month, every single day item (1-31) will be "before" the month footer at the bottom.

Sources/Internal/LayoutItem.swift
```diff
@@ -66,13 +70,26 @@ extension LayoutItem.ItemType: Comparable {
...
+
+    case let (.monthFooter(lhsMonth), .monthHeader(rhsMonth)):
+      return lhsMonth <= rhsMonth
+    case let (.monthFooter(lhsMonth), .dayOfWeekInMonth(_, rhsMonth)):
+      return lhsMonth < rhsMonth
+    case let (.monthFooter(lhsMonth), .day(rhsDay)):
+      return lhsMonth <= rhsDay.month
+    case let (.monthFooter(lhsMonth), .monthFooter(rhsMonth)):
+      return lhsMonth < rhsMonth
```
bryankeller, line 92:
```suggestion
    case let (.monthFooter(lhsMonth), .monthHeader(rhsMonth)):
      return lhsMonth < rhsMonth
    case let (.monthFooter(lhsMonth), .dayOfWeekInMonth(_, rhsMonth)):
      return lhsMonth < rhsMonth
    case let (.monthFooter(lhsMonth), .day(rhsDay)):
      return lhsMonth < rhsDay.month
    case let (.monthFooter(lhsMonth), .monthFooter(rhsMonth)):
      return lhsMonth < rhsMonth
```

Now the inverse fix - a footer is only considered "less than" another item in the calendar if that other item is in a month in the future. 

If the other item is in the _same month_ as the footer on the lefthand side of the comparison `<`, then there's no way that the footer can be "less than" that other item, since the footer is the last item in the month. In other words, the footer is the last item in each month, so it can't be less than any other item in that same month.

Sources/Internal/LayoutItemTypeEnumerator.swift
```diff
@@ -127,13 +132,16 @@ final class LayoutItemTypeEnumerator {
...
+        return .monthFooter(day.month)
       } else if day == dayRange.upperBound {
         let nextMonth = calendar.month(byAddingMonths: 1, to: nextDay.month)
         return .monthHeader(nextMonth)
       } else {
         return .day(nextDay)
       }
+    
+    case .monthFooter(let month):
+      return .monthHeader(calendar.month(byAddingMonths: 1, to: month))
```
bryankeller, line 144:
```suggestion
      let nextMonth = calendar.month(byAddingMonths: 1, to: month)
      return .monthHeader(nextMonth)
```

For consistency with the rest of the cases ^

Sources/Public/CalendarViewContent.swift
```diff
@@ -323,6 +350,7 @@ public final class CalendarViewContent {
 
   // TODO(BK): Make all item provider closures private(set) after legacy `CalendarItem` is removed.
   var monthHeaderItemModelProvider: (Month) -> InternalAnyCalendarItemModel
+  var monthFooterItemModelProvider: (Month) -> InternalAnyCalendarItemModel?
```
bryankeller, line 353: small nit: can we make this entire closure optional instead of just the return value?

```suggestion
  var monthFooterItemModelProvider: ((Month) -> InternalAnyCalendarItemModel)?
```

Sources/Public/CalendarViewContent.swift
```diff
@@ -93,6 +93,11 @@ public final class CalendarViewContent {
         dateFormatter: dayDateFormatter)
       return .itemModel(itemModel)
     }
+    
+    monthFooterItemModelProvider = { month in
+      return nil
+    }
+
```
bryankeller, line 100:
```suggestion
```

Won't be needed after the entire closure is made to be optional

Sources/Public/CalendarView.swift
```diff
@@ -545,6 +548,30 @@ public final class CalendarView: UIView {
...
+    
+  private func monthFooterHeight() -> CGFloat {
+    let monthWidth: CGFloat
+    switch content.monthsLayout {
+    case .vertical:
+      monthWidth = bounds.width
+    case .horizontal(let _monthWidth): monthWidth = _monthWidth
+    }
+    
+    let firstMonthFooterItemModel = content.monthFooterItemModelProvider(
```
bryankeller, line 560:
```suggestion
    let firstMonthFooterItemModel = content.monthFooterItemModelProvider?(
```

Sources/Public/CalendarView.swift
```diff
@@ -545,6 +548,30 @@ public final class CalendarView: UIView {
       verticalFittingPriority: .fittingSizeLevel)
     return size.height
   }
+    
+  private func monthFooterHeight() -> CGFloat {
```
bryankeller, line 552:
```suggestion
  private func monthFooterHeight() -> CGFloat? {
```

Sources/Public/CalendarView.swift
```diff
@@ -545,6 +548,30 @@ public final class CalendarView: UIView {
...
+      let size = firstMonthFooter.systemLayoutSizeFitting(
+        CGSize(width: monthWidth, height: 0),
+        withHorizontalFittingPriority: .required,
+        verticalFittingPriority: .fittingSizeLevel)
+      return size.height
+      
+    }
+    
+    return 0.0
+  }
```
bryankeller, line 574:
```suggestion
  private func monthFooterHeight() -> CGFloat {
    guard 
      let firstMonthFooterItemModel = content.monthFooterItemModelProvider?(
        content.monthRange.lowerBound)
    else
    {
      return nil
    }

    let monthWidth: CGFloat
    switch content.monthsLayout {
    case .vertical: monthWidth = bounds.width
    case .horizontal(let _monthWidth): monthWidth = _monthWidth
    }
    
    firstMonthFooterItemModel.setViewModelOnViewOfSameType(firstMonthFooter)
      
    let size = firstMonthFooter.systemLayoutSizeFitting(
      CGSize(width: monthWidth, height: 0),
      withHorizontalFittingPriority: .required,
      verticalFittingPriority: .fittingSizeLevel)
    return size
  }
```

Sources/Public/CalendarViewContent.swift
```diff
@@ -194,6 +199,28 @@ public final class CalendarViewContent {
...
   }
+    
+  /// Configures the month footer item provider.
+  ///
+  /// `CalendarView` invokes the provided `monthFooterItemModelProvider` for each month in the range of months being
+  /// displayed. The `CalendarItemModel`s that you return will be used to create the views for each month footer in
+  /// `CalendarView`.
+  ///
+  /// If you don't configure your own month header item provider via this function, then a default month header item provider will be
+  /// used.
```
bryankeller, line 210:
```suggestion
  /// If you don't configure a month footer item provider via this function, then no month footers will be displayed.
```

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -277,8 +279,8 @@ final class VisibleItemsProvider {
       case .day(let day):
         month = day.month
         calendarItemModel = self.content.dayItemModelProvider(day)
-      case .dayOfWeekInMonth:
-        return
+      case .dayOfWeekInMonth: return
+      case .monthFooter: return
```
bryankeller, line 283:
```suggestion
      case .dayOfWeekInMonth:
        return
      case .monthFooter(let _month):
        month = _month
        if let monthFooterItemModelProvider = self.content.monthFooterItemModelProvider(month) {
          calendarItemModel = monthFooterItemModelProvider
        } else {
          return
        }
```

Without this, we'll omit footers for people using VoiceOver and other accessibility technology.

Sources/Public/CalendarViewContent.swift
```diff
@@ -194,6 +199,28 @@ public final class CalendarViewContent {
...
+  ///
+  /// - Parameters:
+  ///   - monthFooterItemModelProvider: A closure (that is retained) that returns a `CalendarItemModel` representing a
+  ///   month footer.
+  ///   - month: The `Month` for which to provide a month footer item.
+  /// - Returns: A mutated `CalendarViewContent` instance with a new month footer item provider.
+  public func withMonthFooterItemModelProvider(
+    _ monthFooterItemModelProvider: @escaping (_ month: Month) -> AnyCalendarItemModel)
```
bryankeller, line 218:
```suggestion
    _ monthFooterItemModelProvider: @escaping (_ month: Month) -> AnyCalendarItemModel?)
```

Sources/Public/CalendarViewContent.swift
```diff
@@ -194,6 +199,28 @@ public final class CalendarViewContent {
     self.monthHeaderItemModelProvider = { .itemModel(monthHeaderItemModelProvider($0)) }
     return self
   }
+    
+  /// Configures the month footer item provider.
+  ///
+  /// `CalendarView` invokes the provided `monthFooterItemModelProvider` for each month in the range of months being
+  /// displayed. The `CalendarItemModel`s that you return will be used to create the views for each month footer in
+  /// `CalendarView`.
```
bryankeller, line 207:
```suggestion
  /// `CalendarView`. If you return `nil` for a particular month, then that month will not contain a month footer.
```

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -614,7 +618,7 @@ final class VisibleItemsProvider {
 
         let itemType = VisibleCalendarItem.ItemType.layoutItemType(layoutItem.itemType)
 
-        let calendarItemModel: InternalAnyCalendarItemModel
+        var calendarItemModel: InternalAnyCalendarItemModel? = nil
```
bryankeller, line 621:
```suggestion
        let calendarItemModel: InternalAnyCalendarItemModel?
```

Let's keep this as a `let` so that it's clear that it can only be set once.

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -691,13 +695,25 @@ final class VisibleItemsProvider {
...
+        
+        // We only want to add the visible item if we have one.
+        // Footers are optional, so this check is now necessary.
+        
+        if let itemModel = calendarItemModel {
+          let visibleItem = VisibleCalendarItem(
+            calendarItemModel: itemModel,
+            itemType: .layoutItemType(layoutItem.itemType),
+            frame: layoutItem.frame)
+          visibleItems.insert(visibleItem)
```
bryankeller, line 715:
```suggestion
        case .monthFooter(let month):
          if let footerProvider = content.monthFooterItemModelProvider(month) {
            calendarItemModel = calendarItemModelCache.value(
              for: itemType,
              missingValueProvider: {
                previousCalendarItemModelCache?[itemType] ?? footerProvider
              })
          } else {
            calendarItemModel = nil
          }
        }
        
        if let itemModel = calendarItemModel {
          let visibleItem = VisibleCalendarItem(
            calendarItemModel: itemModel,
            itemType: .layoutItemType(layoutItem.itemType),
            frame: layoutItem.frame)
          visibleItems.insert(visibleItem)
```

Sources/Public/CalendarViewContent.swift
```diff
@@ -194,6 +199,28 @@ public final class CalendarViewContent {
...
+  /// - Parameters:
+  ///   - monthFooterItemModelProvider: A closure (that is retained) that returns a `CalendarItemModel` representing a
+  ///   month footer.
+  ///   - month: The `Month` for which to provide a month footer item.
+  /// - Returns: A mutated `CalendarViewContent` instance with a new month footer item provider.
+  public func withMonthFooterItemModelProvider(
+    _ monthFooterItemModelProvider: @escaping (_ month: Month) -> AnyCalendarItemModel)
+    -> CalendarViewContent
+  {
+    self.monthFooterItemModelProvider = { .itemModel(monthFooterItemModelProvider($0)) }
```
bryankeller, line 221:
```suggestion
    self.monthFooterItemModelProvider = {
      guard let footerItemModel = monthFooterItemModelProvider($0) else { return nil }
      return .itemModel(footerItemModel)
    }
```

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -691,13 +695,25 @@ final class VisibleItemsProvider {
           if framesForVisibleDays[day] == nil {
             framesForVisibleDays[day] = layoutItem.frame
           }
+            
+        case .monthFooter(let month):
+          if let footerProvider = content.monthFooterItemModelProvider(month) {
+            calendarItemModel = calendarItemModelCache.value(for: itemType, missingValueProvider: {
+              previousCalendarItemModelCache?[itemType] ?? footerProvider
+            })
+          }
```
bryankeller, line 704:
```suggestion
        case .monthFooter(let month):
          calendarItemModel = content.monthFooterItemModelProvider?(month)
```

bryankeller:
@saudeon this is looking really good / close. I added a bunch of code suggestions, hopefully it makes things easy since you can commit them right from GitHub. Let me know if you'd like to discuss any further, happy to chat!

saudeon:
@bryankeller going to spend some time going over this today so that we can get it merged in ASAP. Been way too busy with _actual_ work and personal projects, so my apologies!

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/MonthFootersDemoViewController.swift
saudeon: Agreed!

Sources/Internal/LayoutItem.swift
saudeon: Good point, I think this was leftovers from how I first positioned these!

Sources/Internal/LayoutItemTypeEnumerator.swift
```diff
@@ -127,13 +132,17 @@ final class LayoutItemTypeEnumerator {
     case .day(let day):
       let nextDay = calendar.day(byAddingDays: 1, to: day)
       if day.month != nextDay.month {
-        return .monthHeader(nextDay.month)
+        return .monthFooter(day.month)
```
bryankeller, line 135: Should this actually be combined with the `else if day == dayRange.upperBound` (which would indicate that we're at the last day in the total day-range of the calendar? If we don't return a monthFooter from that case, then we'd potentially miss showing a month footer at the end of the calendar.

bryankeller:
> @bryankeller going to spend some time going over this today so that we can get it merged in ASAP. Been way too busy with _actual_ work and personal projects, so my apologies!

Totally get that 😄 This looks great right now - I'm guessing the big thing missing right now is unit tests. Unfortunately, that's going to be a bit of a pain since the `VisibleItemsProvider` tests are currently just using frames from printing out a description of the layout. If you add a footer to the layout, then all of those hard-coded frames are going to be offset by the wrong amount.

You might be able to make your life easier if you set the footer height to something like 10, which will allow you to go through and just add 10 to all of the subsequent item's y positions. It's still going to be a bit of a pain but might be easier than regenerating the tests from scratch. If you'd like some help with this part, I'd be happy to take a look to see how much effort it will be.

saudeon:
> Totally get that 😄 This looks great right now - I'm guessing the big thing missing right now is unit tests. Unfortunately, that's going to be a bit of a pain since the `VisibleItemsProvider` tests are currently just using frames from printing out a description of the layout. If you add a footer to the layout, then all of those hard-coded frames are going to be offset by the wrong amount.
> 
> You might be able to make your life easier if you set the footer height to something like 10, which will allow you to go through and just add 10 to all of the subsequent item's y positions. It's still going to be a bit of a pain but might be easier than regenerating the tests from scratch. If you'd like some help with this part, I'd be happy to take a look to see how much effort it will be.

@bryankeller I have no idea how I missed this comment!

I can run through and modify the tests by adjusting the footer height, although the tests utilizing frame sizes etc could deem quite cumbersome. I wonder if it is worth re-thinking how we can unit test the layout occurring in a way that is extensible and scalable?

bryankeller:
> I can run through and modify the tests by adjusting the footer height, although the tests utilizing frame sizes etc could deem quite cumbersome. I wonder if it is worth re-thinking how we can unit test the layout occurring in a way that is extensible and scalable?

I definitely think this is something that needs to be done. I wonder how other projects test their layouts / frames. If you have any ideas would love to think things through!

----

In the mean time, maybe we can make all of the footer unit tests separate, that way the existing `VisibleItemsProvider` tests don't need to be updated.

Mordil:
Sorry to revive this PR, but what happened to the feature, ultimately?

I see that the original feature request in #31 is open, and the only PR I can find is this one.

Was it just a fact of unit tests being missing before it could be merged?

PR #79, opened 2020-12-01
Bk/reduce frame math duplication

bryankeller:
## Details

This consolidates similar math / calculations in the `FrameProvider`. There were several places where I was determining the height of a certain number of day rows (weeks in a month) the same way, so it seems like a good place to consolidate the math into some private helper functions.

## Related Issue

N/A

## Motivation and Context

Simplify frame calculation logic since I noticed that it was tricky to get some of this stuff right in the Month Footer PR (#75)

## How Has This Been Tested

Unit tests, example app.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Internal/FrameProvider.swift
+++ b/Sources/Internal/FrameProvider.swift
@@ -140,13 +140,15 @@ final class FrameProvider {
     let date = calendar.startDate(of: day)
     let dayOfWeekPosition = calendar.dayOfWeekPosition(for: date)
     let rowInMonth = adjustedRowInMonth(for: day)
+    let numberOfWeekRows = rowInMonth + 1
 
     let x = minXOfItem(at: dayOfWeekPosition, minXOfContainingRow: monthOrigin.x)
     let y = monthOrigin.y +
       monthHeaderHeight +
       content.monthDayInsets.top +
-      (monthsLayout.pinDaysOfWeekToTop ? 0 : (daySize.height + content.verticalDayMargin)) +
-      (CGFloat(rowInMonth) * (daySize.height + content.verticalDayMargin))
+      heightOfDaysOfTheWeekRowInMonth() +
+      heightOfDayContent(forNumberOfWeekRows: numberOfWeekRows) -
+      daySize.height
     return CGRect(origin: CGPoint(x: x, y: y), size: daySize)
   }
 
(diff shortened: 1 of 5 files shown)


Tests/FrameProviderTests.swift
```diff
@@ -171,7 +171,7 @@ final class FrameProviderTests: XCTestCase {
         itemType: .dayOfWeekInMonth(
           position: .fourth,
           month: Month(era: 1, year: 2020, month: 01, isInGregorianCalendar: true)),
-        frame: CGRect(x: 130, y: 200, width: 40, height: 40))
```
bryankeller, line 174: the made up 40x40 size used to work because the calculation in FrameProvider was based on `frame.minY`. Now that the calculation is based on `frame.maxY`, we can't just use fake size values. Instead, this should be the real day size from `frameProvider.daySize`, otherwise the resulting calculations will be nonsense.

Sources/Internal/FrameProvider.swift
```diff
@@ -362,4 +363,20 @@ final class FrameProvider {
...
+  // the `content.verticalDayMargin`.
+  private func heightOfDayContent(forNumberOfWeekRows numberOfWeekRows: Int) -> CGFloat {
+    (CGFloat(numberOfWeekRows) * daySize.height) +
+      (CGFloat(numberOfWeekRows - 1) * content.verticalDayMargin)
+  }
+
+  // Gets the height of the days of the week row, plus the padding between it and the first row of
+  // days in each month. The returned value will be `0` if
+  // `monthsLayout.pinDaysOfWeekToTop == true`.
+  private func heightOfDaysOfTheWeekRowInMonth() -> CGFloat {
```
saudeon, line 378: Very clean! Nice!

Sources/Internal/FrameProvider.swift
```diff
@@ -140,13 +140,15 @@ final class FrameProvider {
     let date = calendar.startDate(of: day)
     let dayOfWeekPosition = calendar.dayOfWeekPosition(for: date)
     let rowInMonth = adjustedRowInMonth(for: day)
+    let numberOfWeekRows = rowInMonth + 1
```
brynbodayle, line 143:
```suggestion
    let numberOfWeekRowsBeforeDay = rowInMonth + 1
```
Just a thought to improve clarity. At first when reading this I thought it was total number of week rows in the month.

brynbodayle (approved):

Sources/Internal/FrameProvider.swift
bryankeller: Changed the name to `numberOfWeekRowsThroughDay` since it's inclusive (not up-to) the the row the day is in

brynbodayle (dismissed):

PR #80, opened 2020-12-21
Fix floating-point precision layout bug

bryankeller:
## Details

This PR fixes a subtle floating-point precision bug that I noticed when using a horizontal layout. Basically, we'd try comparing something like 1.48999999999 to 1.5, both rounded to the nearest 1/3 pixel on a 3x display. This would result in comparing 1.3333333 to 1.6666666, which would cause them to appear to be not equal even though they're both essentially 1.5.

To fix this, I've introduced a CGFloat comparison function that takes in a threshold parameter and tests for approximate equality using that threshold.

## Related Issue

N/A

## Motivation and Context

Fixes a layout bug.

## How Has This Been Tested

Unit tests, tested on my 3x phone.

## Types of changes

## Checklist


Example/HorizonCalendarExample/HorizonCalendarExample.xcodeproj/project.pbxproj
```diff
@@ -366,7 +366,7 @@
 			buildSettings = {
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CODE_SIGN_STYLE = Automatic;
-				DEVELOPMENT_TEAM = 5Q5SGQT2R4;
+				DEVELOPMENT_TEAM = C835JB7RUE;
```
bryankeller, line 369: new phone, lost my contacts. leave a comment with your number below

Sources/Internal/FrameProvider.swift
```diff
@@ -178,10 +178,7 @@ final class FrameProvider {
     let origin: CGPoint
     if distanceFromAdjacentDay < 0 {
       let proposedX = adjacentDayFrame.minX - content.horizontalDayMargin - daySize.width
-      if
-        proposedX.alignedToPixel(forScreenWithScale: scale) >=
-          minX.alignedToPixel(forScreenWithScale: scale)
```
bryankeller, line 183: it was here that 1.49899999 would round down to 1.333333, and 1.5 would round up to 1.6666666, causing the comparison to fail.

Example/HorizonCalendarExample/HorizonCalendarExample.xcodeproj/project.pbxproj
saudeon: I laughed so hard at this.

brynbodayle (approved):

Issue #90, opened 2021-01-19
Scrolling bug when calendar view extends outside of safe area


hiltonc:
I have the calendar view extending to top and bottom edges of the screen:

```
NSLayoutConstraint.activate([
    calendarView.leadingAnchor.constraint(equalTo: view.layoutMarginsGuide.leadingAnchor),
    calendarView.trailingAnchor.constraint(equalTo: view.layoutMarginsGuide.trailingAnchor),
    calendarView.topAnchor.constraint(equalTo: view.topAnchor),
    calendarView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
])
```

When I scroll to the end it glitches:

```
calendarView.scroll(
            toDayContaining: Date(),
            scrollPosition: .centered,
            animated: true)
```

https://user-images.githubusercontent.com/470753/105067755-e57e7000-5a34-11eb-9b38-f5cb892b3982.mov

If I constrain the top and bottom to the `layoutMarginsGuide` I don't have this issue. I'd prefer to extend to the edges of the screen so that I get the content under the navigation bar and home indicator visual effect.

bryankeller:
Oh yikes! Definitely a bug, thanks for reporting. I'll look into this soon.

bryankeller:
@hiltonc fix will be merged Monday at the latest, thanks for reporting this bug!

https://github.com/airbnb/HorizonCalendar/pull/101

hiltonc:
Awesome, thank you @bryankeller!

bryankeller:
Fixed in v1.7.1 🙂

PR #101, opened 2021-02-06
Bk/fix scroll to item layout margins bug

bryankeller:
## Details

This PR fixes an issue that causes the scroll-to-item animation to get glitchy if there are layout margins set along the same axis as the scroll direction. There was a simple oversight in the target item calculation that resulted in the scroll-to-item being unable to get to its final position.

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/90

## Motivation and Context

Bug fix.

## How Has This Been Tested

Simulator + device.

## Types of changes

## Checklist

### DIFF
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
@@ -50,8 +50,8 @@ class DemoViewController: UIViewController {
     switch monthsLayout {
     case .vertical:
       NSLayoutConstraint.activate([
-        calendarView.topAnchor.constraint(equalTo: view.layoutMarginsGuide.topAnchor),
-        calendarView.bottomAnchor.constraint(equalTo: view.layoutMarginsGuide.bottomAnchor),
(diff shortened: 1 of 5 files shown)


Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
```diff
@@ -50,8 +50,8 @@ class DemoViewController: UIViewController {
     switch monthsLayout {
     case .vertical:
       NSLayoutConstraint.activate([
-        calendarView.topAnchor.constraint(equalTo: view.layoutMarginsGuide.topAnchor),
-        calendarView.bottomAnchor.constraint(equalTo: view.layoutMarginsGuide.bottomAnchor),
+        calendarView.topAnchor.constraint(equalTo: view.topAnchor),
+        calendarView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
```
bryankeller, line 54: now that this bug is fixed, we can constrain to the top / bottom anchor and let the system set our `layoutMargins`. Scroll-to-item will work with no issues.

brynbodayle (dismissed):

PR #102, opened 2021-02-08
Bk/fix layout margins behavior

bryankeller:
## Details

The way I overrode the `layoutMargins` property was wrong. By accessing `self.layoutMargins` in `didSet`, I was actually getting the system-computed layout margins (which includes the additional safe area margins). If someone called `calendarView.layoutMargins = .zero`, `self.layoutMargins.top` in the `didSet` would be `64` instead of `0`.

The correct way to override this is to access `newValue` in `set`. 

## Related Issue

N/A

## Motivation and Context

Make `layoutMargins` work correctly.

## How Has This Been Tested

Simulator

## Types of changes

## Checklist

### DIFF
--- a/Sources/Internal/FrameProvider.swift
+++ b/Sources/Internal/FrameProvider.swift
@@ -207,11 +207,16 @@ final class FrameProvider {
     -> CGRect
   {
     let x = minXOfItem(at: dayOfWeekPosition, minXOfContainingRow: layoutMargins.leading)
-    return CGRect(origin: CGPoint(x: x, y: yContentOffset), size: daySize)
+    let y = layoutMargins.top + yContentOffset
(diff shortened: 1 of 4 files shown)


brynbodayle (approved):

Sources/Internal/FrameProvider.swift
```diff
@@ -221,7 +226,7 @@ final class FrameProvider {
   {
     CGRect(
       x: layoutMargins.leading,
-      y: yContentOffset + daySize.height - separatorHeight,
+      y: layoutMargins.top + yContentOffset + daySize.height - separatorHeight,
```
bryankeller, line 229: the pinned days of the week items also needed to use the top margin, otherwise it'll be obscured by the top nav bar.

PR #114, opened 2021-04-16
Fix view z positions when render in context snapshotting

bryankeller:
## Details

This PR fixes a rendering issue when `CalendarView` is part of a view hierarchy being snapshotted via `CALayer.render(in:)`. The core issue is that, for performance reasons, `CalendarView` does not sort its subviews before adding them to the view hierarchy, and instead simply sets each subview's layer's `zPosition` property. This works well, except when snapshotting a view view `CALayer.render(in:)`. It's unclear if this is a bug with that function or not (it certainly seems like a `CALayer` snapshotting function should respect `sublayer`'s `zPosition`s), but I'm not going to bank on Apple fixing this anytime soon.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 8 - 2021-04-15 at 18 02 17 | Simulator Screen Shot - iPhone 8 - 2021-04-15 at 19 03 23 |

## Related Issue

N/A

## Motivation and Context

Fixes snapshotting for users who want that.

## How Has This Been Tested

Tested in example app by snapshotting demo VC.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -47,6 +47,10 @@ public final class CalendarView: UIView {
 
     super.init(frame: .zero)
 
+    (layer as! RenderInContextObservingLayer).willRenderInContext = { [weak self] in
+      self?.sortSubviewsByLayerZPositions()
+    }
+
     if #available(iOS 13.0, *) {
       backgroundColor = .systemBackground
     } else {
@@ -69,6 +73,8 @@ public final class CalendarView: UIView {
   }
 
   // MARK: Public
+
+  public override class var layerClass: AnyClass { RenderInContextObservingLayer.self }
 
   /// A closure (that is retained) that is invoked whenever a day is selected.
   public var daySelectionHandler: ((Day) -> Void)?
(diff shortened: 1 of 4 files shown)


Sources/Public/CalendarView.swift
```diff
@@ -804,6 +810,22 @@ public final class CalendarView: UIView {
...
+      for j in 0..<(scrollView.subviews.count - i - 1) {
+        let lhsZPosition = scrollView.subviews[j].layer.zPosition
+        let rhsZPosition = scrollView.subviews[j + 1].layer.zPosition
+
+        if lhsZPosition > rhsZPosition {
+          scrollView.exchangeSubview(at: j, withSubviewAt: j + 1)
+        }
+      }
+    }
+  }
```
bryankeller, line 827: So yeah, I could add this as a generic internal internal bubble sort implementation and write unit tests for it... but I kind of don't want to bother 😛

I'm not even sure this is needed long term. I need to profile how expensive it is, but I might be able to just replace my calls to `scrollView.addSubview(view)` with calls to `scrollView.insertSubview(view, at: correctIndex)`. If I can do this, then I don't need to manually set the `zPosition` on sublayers. My main concern with this approach is that I'm doing a lot of subview array insert operations (rather than append operations), which I suspect is more expensive, especially since this is done in a hot codepath while scrolling / views are being reused.

We need a quick fix now for some Airbnb stuff, and this code only ever runs if the layer is snappshotted via `render(in:)`, so trying to keep this lightweight.

brynbodayle (approved):

Issue #104, opened 2021-02-26
Partial visibility doesn't work correctly for < 2 months


Aeroporas:
If I select visibility range in partially visible mode like 

```
let startDate = calendar.date(from: DateComponents(year: 2021, month: 02, day: 5))!
let endDate = calendar.date(from: DateComponents(year: 2021, month: 03, day: 15))!
```

the scrolling function stops working at all and part of the calendar just stucks behind bottom of the screen. The problem is reproducible in example project.

Here is a screenshot

<img width="484" alt="Снимок экрана 2021-02-26 в 17 00 31" src="https://user-images.githubusercontent.com/21102640/109309301-251a5200-7854-11eb-8365-1bf458758ffa.png">

Aeroporas:
When I extend visible dates range all seems to work nice again.

bryankeller:
Oh whoops, definitely a bug. I'll have a fix for this soon. Thanks for the report!

bryankeller:
Thanks for your patience on this @Aeroporas - the fix is in `v1.8.3` 🙂

PR #117, opened 2021-04-21
Bk/fix partial month visibility layout bug

bryankeller:
## Details

This fixes a bug that causes `scrollToMonth/Day` to not work correctly when using partial-month visibility. For example, the calendar can show a subset of dates in the first / last month. In this scenario, the calculations to figure out the programatic scroll were incorrect. I've refactored the code a bit and added some new unit tests.

https://user-images.githubusercontent.com/746571/115615243-951ce580-a2a3-11eb-83e2-a807c33464d8.mov

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/104

## Motivation and Context

Bug fix.

## How Has This Been Tested

Example app, unit tests.

## Types of changes

## Checklist


Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/LargeDayRangeDemoViewController.swift
```diff
@@ -24,12 +24,26 @@ final class LargeDayRangeDemoViewController: DemoViewController {
...
+    switch monthsLayout {
+    case .vertical: padding = calendarView.layoutMargins.top
+    case .horizontal: padding = calendarView.layoutMargins.left
+    }
 
     let january1500CE = calendar.date(from: DateComponents(era: 1, year: 1500, month: 01, day: 01))!
     calendarView.scroll(
       toMonthContaining: january1500CE,
-      scrollPosition: .firstFullyVisiblePosition,
+      scrollPosition: .firstFullyVisiblePosition(padding: padding),
```
bryankeller, line 43: this is just improving the example project behavior a bit. This ensures that the month we scroll to won't be obscured by any bars.

Sources/Internal/FrameProvider.swift
```diff
@@ -243,6 +243,64 @@ final class FrameProvider {
...
+        x = lastFullyVisibleX - padding
+      }
+
+      return CGRect(
+        x: x,
+        y: originalFrame.minY,
+        width: originalFrame.width,
+        height: originalFrame.height)
+    }
+  }
```
bryankeller, line 302: This logic in here is unchanged, but I moved it from `VisibleItemsProvider` to `FrameProvider`, which also makes it easier to unit test.

Sources/Internal/FrameProvider.swift
```diff
@@ -347,7 +412,13 @@ final class FrameProvider {
   // boundary month that's only showing a subset of days.
   private func numberOfWeekRows(in month: Month) -> Int {
     let rowOfLastDateInMonth: Int
-    if month == content.monthRange.lowerBound {
+    if month == content.monthRange.lowerBound && month == content.monthRange.upperBound {
+      let firstDayOfOnlyMonth = content.dayRange.lowerBound
+      let lastDayOfOnlyMonth = content.dayRange.upperBound
+      let rowOfFirstDayOfOnlyMonth = adjustedRowInMonth(for: firstDayOfOnlyMonth)
+      let rowOfLastDayOfOnlyMonth = adjustedRowInMonth(for: lastDayOfOnlyMonth)
+      rowOfLastDateInMonth = rowOfLastDayOfOnlyMonth - rowOfFirstDayOfOnlyMonth
```
bryankeller, line 420: this new case fixes an issue where the calculation for the height of a month would be incorrect if it was both the first and last month in the calendar. This new case handles that properly.

brynbodayle (dismissed):

brynbodayle (approved):

PR #129, opened 2021-07-14
Make accessibility focus declarative

bryankeller:
## Details

Accessibility focus can now be specified declaratively as part of `CalendarViewContent`, rather than needing to invoke `UIAccessibility.post(...)` from feature code with the correct accessibility element (although the API to do it that way still exists, but is marked as deprecated).

## Related Issue

N/A

## Motivation and Context

Noticing some bugs in the Airbnb app where the default a11y selection is not correct. Adding support for this should make it much easier to solve those bugs. Even the included example project had some issues with a11y focus changing unexpectedly after an item was selected.

## How Has This Been Tested

Tested on real device with VoiceOver running.

## Types of changes

## Checklist


Sources/Public/CalendarView.swift
```diff
@@ -247,7 +247,7 @@ public final class CalendarView: UIView {
     cachedAccessibilityElements = nil
     if let element = focusedAccessibilityElement as? OffScreenCalendarItemAccessibilityElement {
       UIAccessibility.post(
-        notification: .layoutChanged,
+        notification: .screenChanged,
```
bryankeller, line 250: using `.screenChanged` seems to result in the focus programmatically changing faster. Using `.layoutChanged` results in a delay.

Sources/Public/CalendarView.swift
```diff
@@ -659,11 +665,20 @@ public final class CalendarView: UIView {
...
-    // Set up the selection handler
+    // Setup specific to day items
     if case .layoutItemType(.day(let day)) = visibleItem.itemType {
       view.selectionHandler = { [weak self] in
         self?.daySelectionHandler?(day)
       }
+
+      if day == content.dayForAccessibilityFocusOverride {
+        // Without this delay, VoiceOver won't respect the focused element change on first load.
+        // After first load, this delay is not needed.
```
bryankeller, line 676: image

bryankeller, line 676: giphy

Sources/Public/CalendarView.swift
brynbodayle: ah good catch, this looks more correct based on the docs too

Sources/Public/CalendarViewContent.swift
```diff
@@ -173,6 +173,21 @@ public final class CalendarViewContent {
...
 
+  /// Configures the accessibility focus override to be on a particular day after `CalendarView`'s content is set. If you'd like
+  /// VoiceOver to focus on a particular day (e.g. a day that your feature code considers to be selected), then invoke this function with a
+  /// date representing the target day.
+  ///
+  /// - Parameters:
+  ///   - date: The day to be focused, specified with a `Date` instance which will be used to determine the associated day
+  /// using the `calendar` instance with which `CalendarViewContent` was instantiated.
+  /// - Returns: A mutated `CalendarViewContent` instance with a new accessibility focus override.
+  public func withAccessibilityFocusOverride(onDayContaining date: Date?) -> CalendarViewContent {
```
brynbodayle, line 184:
```suggestion
  public func withAccessibilityInitialFocusOverride(onDayContaining date: Date?) -> CalendarViewContent {
```
Might be helpful to use the term `initial` here

Sources/Public/CalendarView.swift
```diff
@@ -247,7 +247,7 @@ public final class CalendarView: UIView {
     cachedAccessibilityElements = nil
     if let element = focusedAccessibilityElement as? OffScreenCalendarItemAccessibilityElement {
       UIAccessibility.post(
-        notification: .layoutChanged,
+        notification: .screenChanged,
         argument: visibleViewsForVisibleItems[element.correspondingItem])
```
brynbodayle, line 251: I'm confused by the two paths we have to focus items. This one in `layoutSubviews` is for visible views but we check if it conforms to `OffScreenCalendarItemAccessibilityElement` first?

And then why are we focusing separately in the configure function?

brynbodayle (commented):
A few questions/comments

bryankeller:
Thinking I'm going to table this idea for now. `scrollToDay` / `scrollToMonth` are also handled imperatively, and one of the motivations for that was that it might feel unnatural for developers to regenerate their content whenever they want to change the scroll position (even if none of their underlying data has changed). There's probably an argument to be made for having every single aspect of the calendar's state driven by `CalendarViewContent` (including scroll position, initial a11y focus), but it also seems like having these as separate, imperative functions that can be called will allow consumers to keep the concepts of data updates (changes in availability, date selection, etc) and non-data updates (VO focus, animating to a date or month) separate.

PR #130, opened 2021-07-16
Improve accessibility support in example projects

bryankeller:
## Details

This improves accessibility support in the example project by maintaining the voiceover focus on the last selected day.

## Related Issue

N/A

## Motivation and Context

Show consumers of the library how to handle this common use case.

## How Has This Been Tested

Real device.

## Types of changes

## Checklist


Sources/Public/CalendarView.swift
```diff
@@ -247,7 +247,7 @@ public final class CalendarView: UIView {
     cachedAccessibilityElements = nil
     if let element = focusedAccessibilityElement as? OffScreenCalendarItemAccessibilityElement {
       UIAccessibility.post(
-        notification: .layoutChanged,
+        notification: .screenChanged,
```
bryankeller, line 250: Keeping this change from #129. This scenario happens when a user swipes to an off-screen accessibility element, and triggers the calendar to scroll a bit so that the previously-off-screen elements are now centered in the middle of the calendar's bounds. Since this triggers a significant visual update, `screenChanged` seems appropriate.

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DayRangeSelectionDemoViewController.swift
```diff
@@ -40,6 +40,16 @@ final class DayRangeSelectionDemoViewController: DemoViewController {
       }
 
       self.calendarView.setContent(self.makeContent())
+
+      if
+        UIAccessibility.isVoiceOverRunning,
+        let selectedDate = self.calendar.date(from: day.components)
+      {
+        self.calendarView.layoutIfNeeded()
```
bryankeller, line 48: Forcing layout seems better than using DispatchQueue.main.async/asyncAfter here. It'll only happen when VO is running, and it guarantees that the view will be ready to receive accessibility focus.

brynbodayle (approved):

PR #134, opened 2021-07-26
Preserve scroll position after bounds or layout margins change

bryankeller:
## Details

When the bounds changes (e.g. on rotation) and when the layout margins change, we should attempt to preserve the current scroll position. In the case of layout margins, the following examples explain the new behavior:

1. If a user is scrolled down by several months, changing the top margin won't cause the content to jump. Once they scroll to the top, they'll notice the new top margin.
2. If a user is scrolled to the top already, changing the top margin will cause the topmost content to move down to respect the new top margin.

## Related Issue

N/A

## Motivation and Context

In the Airbnb app, our navigation infrastructure adjusts the top layout margin of the view to account for Epoxy top bars. Without this change, the topmost content would be covered by the top bars until the user manually scrolled to the top. Prior to this change, we would need to constrain `CalendarView`'s top anchor to `view.layoutMarginsGuide.topAnchor`, rather than `view.topAnchor`, which meant that the calendar's content would get clipped when scrolling under translucent top bars.

| Before | After |
| ---- | ---- |
| Simulator Screen Shot - iPhone 8 - 2021-07-26 at 14 17 21 | Simulator Screen Shot - iPhone 8 - 2021-07-26 at 14 15 55 | 

## How Has This Been Tested

Tested in example app and Airbnb app.

## Types of changes

## Checklist

### DIFF
--- a/HorizonCalendar.podspec
+++ b/HorizonCalendar.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |spec|
   spec.name = "HorizonCalendar"
-  spec.version = "1.8.7"
+  spec.version = "1.8.8"
   spec.license = "Apache License, Version 2.0"
   spec.summary = "A declarative, performant, calendar UI component that supports use cases ranging from simple date pickers to fully-featured calendar apps."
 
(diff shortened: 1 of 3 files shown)


bryankeller:
This also allows us to use `CalendarView` as the view of a view controller (by overriding loadView / without adding it as a subview to an empty view).

brynbodayle (approved):
good stuff!

Issue #89, opened 2021-01-11
Day of Week background 


SaitamLe:
Is there a way to change the background color of the days of week section when those are pinned to the top of the calendar?
 
<img width="409" alt="Bildschirmfoto 2021-01-11 um 10 07 47" src="https://user-images.githubusercontent.com/23743401/104162559-1fc48d80-53f5-11eb-9525-afa7d155baf7.png">

bryankeller:
Hi @SaitamLe - sorry for the delay on this. I think this seems like something that we can support. I'll need to add a new API to configure this on CalendarViewContent, but it should be pretty straightforward. I'll get a change up this week for it.

iosdevmax:
Looking forward to this change as well :)

haffez23:
> und color of the days of week section when those are pinned to the top of the calendar

How did u display three characters in days of week ?

Mordil:
@haffez23 

Unfortunately your only option right now is to just fully override the view with your own implementation with `withDayOfWeekItemModelProvider` which provides `Month?, Int` parameters with the 2nd being the `dayOfWeekIndex` that you can index into `Calendar` to get which day it is and format the value how you want

bryankeller:
Thinking this through a bit more, it seems like right now there are 2 options:

- Use the default items, which don't allow any customization
- Provide your own custom items, which allow full customization

I think there's a middle-ground here, where a customization point is provided for the default items. I agree it seems silly to need to create your own custom items just to change a font size or a background color. Will give this some thought in the next few days.

Mordil:
@bryankeller When I was going to implement my header, what I really wanted was to just access and provide an instance of the default's `InvariantViewProperties` but the entire struct's hierarchy is internal

bryankeller:
@Mordil makes total sense! My current plan is to just expose these publicly. For the default `DayView`, I'm just going to open source the one I built for Airbnb internally, which has iPad pointer interaction support, configurable haptics, and more 🙂

bryankeller:
Hey all, I've got a PR out which exposes configurable default views for month headers, days of the week, and individual days https://github.com/airbnb/HorizonCalendar/pull/140

Should be merged in the next few days - hopefully you can delete some code on your end once this lands! 🤞

bryankeller:
Merged! `v1.10.0` contains a public `DayView`, `DayOfWeekView` and `MonthHeaderView`.

PR #140, opened 2021-09-12
Bk/improve and expose default views

bryankeller:
## Details

Prior to this change, you had 2 choices:

- Use the simple, non-configurable default items
- Provide your own views via `CalendarViewContent`'s provider functions / closures (needed for even the most minor visual tweaks)

This PR addresses the issue of not being able to tweak the default views. To do this, I've:

- Exposed the default views as `public`
- Made them highly configurable via their `InvariantViewProperties`

The (now `public`) `DayView` is the real star here. It has a ton of configurable features like:
- Font, text alignment, text color
- Configurable interaction modes (disabled or enabled)
- A background layer / shape, which can be used to style the view to look selected or just give each view a subtle backdrop
- A highlight layer / shape, which will appear on touch down if interaction is enabled
- Configurable haptic feedback and iPad pointer interaction support
- Configurable accessibility traits (`button` vs. `staticText`, for example)

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/89

## Motivation and Context

Make a lot of common use cases way simpler.

## How Has This Been Tested

I took screenshots of the new default items and compared it with an old screenshot, and they're pixel accurate, so there should be no visual regressions.

## Types of changes

## Checklist


Sources/Public/Item Views/DayOfWeekView.swift
```diff
@@ -0,0 +1,218 @@
...
+
+    super.init(frame: .zero)
+
+    isUserInteractionEnabled = false
+
+    backgroundColor = invariantViewProperties.backgroundColor
+
+    layer.addSublayer(backgroundLayer)
+
+    label.translatesAutoresizingMaskIntoConstraints = true
```
brynbodayle, line 47: I'd remove this since we're not using constraints and this already defaults to true. Either value wouldn't affect this component.

Sources/Public/Item Views/Shape.swift
```diff
@@ -0,0 +1,24 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import CoreGraphics
+
+/// Represents a background or highlight layer shape; used by `DayView` and `DayOfWeekView`.
+public enum Shape: Hashable {
+  case circle
+  case roundRect(cornerRadius: CGFloat)
```
brynbodayle, line 21:
```suggestion
  case rectangle(cornerRadius: CGFloat)
```
since it's not always a rounded rect

Sources/Public/Item Views/DayView.swift
```diff
@@ -0,0 +1,344 @@
...
+  private let invariantViewProperties: InvariantViewProperties
+  private let backgroundLayer: CAShapeLayer
+  private let highlightLayer: CAShapeLayer?
+  private let label: UILabel
+
+  private var feedbackGenerator: UISelectionFeedbackGenerator?
+
+  private var isHighlighted: Bool {
+    get { !(highlightLayer?.isHidden ?? true) }
+    set { highlightLayer?.isHidden = !newValue }
```
brynbodayle, line 177: In the future some animation here would be nice

brynbodayle (approved):
Some recommendations, looks great!

Sources/Public/Item Views/DayView.swift
bryankeller: Added!

Sources/Public/Item Views/Shape.swift
bryankeller: Renamed to `square` since day / day-of-the-week views are always 1:1 aspect ratio.

Sources/Public/Item Views/DayView.swift
bryankeller: Commit for it is https://github.com/airbnb/HorizonCalendar/pull/140/commits/5d079312654a2e9aa0d4d4a1ea9036b8d97c36dd

Sources/Public/Item Views/Shape.swift
brynbodayle: Seems like a requirement that might change independent of this API though.

brynbodayle (approved):

Sources/Public/Item Views/Shape.swift
bryankeller: That's true. Alright, back to `rectangle` it is!

brynbodayle (approved):

Issue #125, opened 2021-05-11
iPad is not supported


isaacthedeveloper:
When I try to run this on the iPad I get the following error. is Ipad supported? 
ERROR:
Thread 1: Assertion failed:   Calendar metrics and size resulted in a negative-or-zero amount of remaining height
  (-460.5 points) after allocating room for calendar elements. If ignored, this
  will cause very odd / incorrect layouts
Assert Code:
```
assert(availableHeight > 0,
        """
          Calendar metrics and size resulted in a negative-or-zero amount of remaining height
          (\(availableHeight) points) after allocating room for calendar elements. If ignored, this
          will cause very odd / incorrect layouts.
        """)
    }
```

isaacthedeveloper:
@bryankeller

bryankeller:
@isaacthedeveloper a temporary workaround would be to increase the available height of `CalendarView`. I agree that this behavior is super fragile and makes working with horizontal calendars really difficult. I think that the ideal behavior would be to have the calendar's height be determined based on the width + metrics that you set. I'll look into making this improvement soon.

bryankeller:
Starting in version 1.11.0 (just released), `CalendarView` now has an `intrinsicContentSize.height` equal to the exact amount of space needed to display 1 month, given the current width / other layout metrics. That means that you can add a horizontally-scrolling `CalendarView` to your view hierarchy, constrain its width, and the height will be set based on the `intrinsicContentSize`.

Issue #110, opened 2021-03-26
Adjusting the size of the calendar


say-when:
I've got a question. I've been evaluating the calendar component for a project, and keep getting an assert "Calendar metrics and size resulted in a negative-or-zero amount of remaining height" when placing it in a container view (about 375 x 375 points). I have to size it up to about 385 height to get it to appear. I couldn't find anything in the docs about sizing the calendar, the day items or adjusting the layout beyond the CalendarViewContent .with methods. Is there something I'm missing?

bryankeller:
Hi @say-when - I know that this is a frustrating behavior. At a minimum, the error that's printed should be improved to explain why the issue is occurring. I also think we could improve the docs for this - I'll do that soon.

If you must fit your calendar in a 375x375 container, you'll need to increase the number of months visible using the horizontal layout config. By displaying more months horizontally, each month will have smaller day items to fit the width, which will also decrease the height, allowing you to fit your calendar in the 375pt tall container.

Try doing this:

```swift
CalendarView(
  visibleDateRange: startDate...endDate,
  monthsLayout: .horizontal(options: HorizontalMonthsLayoutOptions(maximumFullyVisibleMonths: 2)))
```

Try changing the `maximumFullyVisibleMonths` until the content fits correctly. Hope this helps!

bryankeller:
Starting in version 1.11.0 (just released), `CalendarView` now has an `intrinsicContentSize.height` equal to the exact amount of space needed to fit all of the rows of a month, given the current width / other layout metrics. That means that you can add a horizontally-scrolling `CalendarView` to your view hierarchy, constrain its width, and the height will be set based on the `intrinsicContentSize`.

Ksen17:
Version 1.11.0
<img width="401" alt="image" src="https://user-images.githubusercontent.com/8882963/161248843-dcaae907-9467-4580-b1b0-8c91215ea6cf.png">

Issue #88, opened 2021-01-05
Question: Get the height of the displayed content


cuongtvwru:
Hi all, 
How can I get the height of the displayed content?. In my case I display the current month, thanks

bryankeller:
Hi @cuongtvwru - there's not currently an API to do this. Can you explain your use-case a bit more?

bryankeller:
Starting in version 1.11.0 (just released), `CalendarView` now has an `intrinsicContentSize.height` equal to the exact amount of space needed to fit all of the rows of a month, given the current width / other layout metrics. That means that you can add a horizontally-scrolling `CalendarView` to your view hierarchy, constrain its width, and the height will be set based on the `intrinsicContentSize`.

If you want to get the height if the displayed content, simply check `calendarView.bounds.height` after your view hierarchy has been completely laid out.

PR #150, opened 2021-11-03
Bk/make horizontal calendars easier

bryankeller:
#### Background

The PR makes it easier to implement horizontally-scrolling calendars. The really annoying issue before was that `HorizonCalendar` would assert if the available height of the calendar was too small to fit the calendar's contents. In practice, this was a pretty easy scenario to hit. Consider the following:

1. Constrain a `CalendarView` instance to have a height of 200pt and a width of 375pt (wider than it is tall)
2. Use `MonthsLayout.horizontal` with `maximumFullyVisibleMonths` set to 1.
3. Assert :(

The problem is that the size of each day in a horizontally-scrolling calendar is determined by looking at the available width (375pt) and the `maximumFullyVisibleMonths` (1). If we only want to display 1 month in the available width, then that's going to result in some pretty large day items, which means each row of days in a month is going to be pretty tall (day views are 1:1 aspect ratio so big width = big height).

Here's what that scenario looks like (if I ignore the assert):

As we can see, a large portion of the month is getting clipped. What's worse is that API consumers have no way of knowing what to set their `CalendarView`'s `frame.height` to in order to prevent clipping / the assert.

#### Solution

This PR makes 2 improvements:
- Get rid of the pesky assert, and just let the calendar's content get clipped
- Use a sizing `UILabel` to take advantage of `-[UIView _needsDoubleUpdateConstraintPass]` without actually calling that private function ourselves. We know this method works since we use it in the Airbnb app. The result is that `intrinsicContentSize` is invoked multiple times, allowing CalendarView to return an `intrinsicContentSize` with an intrinsic height that depends on a known width.

| Vertical Intrinsic Height | Horizontal Intrinsic Height 1 | Horizontal Intrinsic Height 2 |
| ---- | ---- | ---- |
| Simulator Screen Recording - iPhone 12 Pro - 2021-11-06 at 01 14 26 | Simulator Screen Recording - iPhone 12 Pro - 2021-11-06 at 01 15 34 | Simulator Screen Recording - iPhone 12 Pro - 2021-11-06 at 01 16 23 |

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/88
https://github.com/airbnb/HorizonCalendar/issues/110
https://github.com/airbnb/HorizonCalendar/issues/125

## Motivation and Context

Several people have struggled with this. I think we'd struggle with it too if we used horizontal-scrolling calendars anywhere in the Airbnb app. Anyways, I've put off a fix for this long enough. This should unblock people!

## How Has This Been Tested

Tested iPhone and iPad, added unit tests

## Types of changes

## Checklist

### DIFF
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
@@ -35,6 +35,12 @@ class DemoViewController: UIViewController {
     return dateFormatter
   }()
 
+  // We use a placeholder height until `calendarView.maximumMonthHeightDidChange` is invoked. Once
+  // we know the maximum size that a month can be in the horizontally-scrolling calendar, we can
+  // adjust this constraint's constant so that it perfectly fits each month.
+  lazy var horizontalCalendarViewHeightConstraint = calendarView.heightAnchor.constraint(
+    equalToConstant: 0)
+
   override func viewDidLoad() {
     super.viewDidLoad()
 
@@ -60,15 +66,21 @@ class DemoViewController: UIViewController {
         calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375)
       ])
     case .horizontal:
+      calendarView.maximumMonthHeightDidChange = { [weak self] maximumMonthHeight in
+        guard let self = self else { return }
+        let heightConstant = maximumMonthHeight +
+          self.calendarView.layoutMargins.top +
+          self.calendarView.layoutMargins.bottom
+        self.horizontalCalendarViewHeightConstraint.constant = heightConstant
+      }
+
       NSLayoutConstraint.activate([
         calendarView.centerYAnchor.constraint(equalTo: view.layoutMarginsGuide.centerYAnchor),
-        calendarView.heightAnchor.constraint(equalToConstant: 275),
-        calendarView.leadingAnchor.constraint(
-          greaterThanOrEqualTo: view.leadingAnchor),
-        calendarView.trailingAnchor.constraint(
-          lessThanOrEqualTo: view.trailingAnchor),
+        calendarView.leadingAnchor.constraint(greaterThanOrEqualTo: view.leadingAnchor),
+        calendarView.trailingAnchor.constraint(lessThanOrEqualTo: view.trailingAnchor),
         calendarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
-        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375)
+        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375),
+        horizontalCalendarViewHeightConstraint,
       ])
     }
   }
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
@@ -35,12 +35,6 @@ class DemoViewController: UIViewController {
     return dateFormatter
   }()
 
-  // We use a placeholder height until `calendarView.maximumMonthHeightDidChange` is invoked. Once
-  // we know the maximum size that a month can be in the horizontally-scrolling calendar, we can
-  // adjust this constraint's constant so that it perfectly fits each month.
-  lazy var horizontalCalendarViewHeightConstraint = calendarView.heightAnchor.constraint(
-    equalToConstant: 0)
-
   override func viewDidLoad() {
     super.viewDidLoad()
 
@@ -63,30 +57,34 @@ class DemoViewController: UIViewController {
         calendarView.trailingAnchor.constraint(
           lessThanOrEqualTo: view.layoutMarginsGuide.trailingAnchor),
         calendarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
-        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375)
+        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375),
+        calendarView.widthAnchor.constraint(equalToConstant: 375).prioritize(at: .defaultLow),
       ])
     case .horizontal:
-      calendarView.maximumMonthHeightDidChange = { [weak self] maximumMonthHeight in
-        guard let self = self else { return }
-        let heightConstant = maximumMonthHeight +
-          self.calendarView.layoutMargins.top +
-          self.calendarView.layoutMargins.bottom
-        self.horizontalCalendarViewHeightConstraint.constant = heightConstant
-      }
-
       NSLayoutConstraint.activate([
         calendarView.centerYAnchor.constraint(equalTo: view.layoutMarginsGuide.centerYAnchor),
         calendarView.leadingAnchor.constraint(greaterThanOrEqualTo: view.leadingAnchor),
         calendarView.trailingAnchor.constraint(lessThanOrEqualTo: view.trailingAnchor),
         calendarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
         calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375),
-        horizontalCalendarViewHeightConstraint,
+        calendarView.widthAnchor.constraint(equalToConstant: 375).prioritize(at: .defaultLow),
       ])
     }
   }
 
   func makeContent() -> CalendarViewContent {
     fatalError("Must be implemented by a subclass.")
+  }
+
+}
+
+// MARK: NSLayoutConstraint + Priority Helper
+
+extension NSLayoutConstraint {
+
+  fileprivate func prioritize(at priority: UILayoutPriority) -> NSLayoutConstraint {
+    self.priority = priority
+    return self
   }
 
 }
--- a/Sources/Internal/DoubleLayoutPassHelpers.swift
+++ b/Sources/Internal/DoubleLayoutPassHelpers.swift
@@ -99,6 +99,7 @@ final class DoubleLayoutPassSizingLabel: UILabel {
     numberOfLines = 0
     isUserInteractionEnabled = false
     isAccessibilityElement = false
+    isHidden = true
   }
 
   @available(*, unavailable)
@@ -118,14 +119,6 @@ final class DoubleLayoutPassSizingLabel: UILabel {
     } else {
       return provider.intrinsicContentSize(forHorizontallyInsetWidth: preferredMaxLayoutWidth)
     }
-  }
-
-  override func draw(_: CGRect) {
-    // Do nothing, this view doesn't render any visual content.
-  }
-
-  override func drawText(in _: CGRect) {
-    // Do nothing, this view doesn't render any visual content.
   }
 
   // MARK: Private
--- a/Sources/Internal/DoubleLayoutPassHelpers.swift
+++ b/Sources/Internal/DoubleLayoutPassHelpers.swift
@@ -1,0 +1,136 @@
+// Created by Bryan Keller on 11/5/21.
+// Copyright © 2021 Airbnb Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import UIKit
+
+// MARK: CalendarView + Double Layout Pass Helpers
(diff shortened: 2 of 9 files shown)


Sources/Internal/ScrollMetricsMutator.swift
```diff
@@ -105,7 +105,7 @@ enum ScrollAxis {
 
 // MARK: - ScrollMetricsProvider
 
-protocol ScrollMetricsProvider: class {
+protocol ScrollMetricsProvider: AnyObject {
```
bryankeller, line 108: just fixing an xcode warning

Sources/Internal/FrameProvider.swift
```diff
@@ -247,7 +256,7 @@ final class FrameProvider {
 
   /// Returns translated item frames for the specified scroll offset and scroll position. Note that the returned frames may not be at valid
   /// resting positions. For example, someone could try to scroll the last day in the calendar to be at a vertically-centered scroll
-  /// position, which would cause the calendar to be laid out in an overscrolled position. The `VisibleItemsProvider` will detect
+  /// position, which would cause the calendar to be laid out in an over-scrolled position. The `VisibleItemsProvider` will detect
```
bryankeller, line 259: typo

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -904,7 +905,7 @@ final class VisibleItemsProvider {
   }
 
   /// This function takes a proposed frame for a target item toward which we're programmatically scrolling, and adjusts it so that it's a
-  /// valid frame when the calendar is at rest / not being overscrolled.
+  /// valid frame when the calendar is at rest / not being over-scrolled.
```
bryankeller, line 908: typo

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -951,9 +952,7 @@ final class VisibleItemsProvider {
     currentMonth = month
     currentMonthFrame = monthFrame
     while
-      bounds.contains(
-        CGPoint(x: currentMonthFrame.maxX - 1, y: currentMonthFrame.maxY - 1)
-          .alignedToPixels(forScreenWithScale: scale)),
```
bryankeller, line 956: 🤮 turns out all I needed was `CGRect.intersects`. I was doing this before because `CGRect(0, 0, 1, 1)` does not contain `CGPoint(1, 1)`. The rect intersects check handles this perfectly though, with no gross origin offsetting.

Sources/Public/CalendarView.swift
```diff
@@ -161,7 +176,7 @@ public final class CalendarView: UIView {
   public override func layoutSubviews() {
     super.layoutSubviews()
 
-    // Setting the scroll view's frame in `layoutSubviews` causes overscrolling to not work. We
+    // Setting the scroll view's frame in `layoutSubviews` causes over-scrolling to not work. We
```
bryankeller, line 179: typo fix

Sources/Public/CalendarView.swift
```diff
@@ -429,7 +451,10 @@ public final class CalendarView: UIView {
   }
 
   private var isReadyForLayout: Bool {
-    bounds.width > 0 && bounds.height > 0
+    // There's no reason to attempt layout unless we have a non-zero `bounds.size`. We'll have a
+    // non-zero size once the `frame` is set to something non-zero, either manually or via the
+    // Auto Layout engine.
+    bounds.size != .zero
```
bryankeller, line 457: Changing this since we need `layoutSubviews` to get far enough so that we hit this line and API consumers can learn about the max month height https://github.com/airbnb/HorizonCalendar/pull/150/files#diff-733b4844a228b4fda6b27e4f88b42a6fc4b04eb589081da658741416a2152bd9R242

If API consumers implement their placeholder height constraint like this `calendarView.heightAnchor.constraint(equalToConstant: 0)`, without this change, `isReadyForLayout` will be `false` and we won't make it far enough in `layoutSubviews` for us to calculate a `maximumMonthHeight` to pass back to API consumers, leaving the `CalendarView` stuck rendering nothing.

I think as long as we have a non-zero `bounds.size`, we have enough info to start computing parts of the layout, which will also allow API consumers to get that `maximumMonthHeightDidChange` even if they used 0 height as their placeholder.

Sources/Public/CalendarView.swift
```diff
@@ -751,7 +776,7 @@ public final class CalendarView: UIView {
     }
   }
   
-  // This hack is needed to prevent the scroll view from overscrolling far past the content. This
+  // This hack is needed to prevent the scroll view from over-scrolling far past the content. This
```
bryankeller, line 779: typoooo

Sources/Public/CalendarView.swift
```diff
@@ -222,6 +237,12 @@ public final class CalendarView: UIView {
         maximumScrollOffset: maximumScrollOffset)
     }
 
+    let maxMonthHeight = currentVisibleItemsDetails.maxMonthHeight
+    if maxMonthHeight != previousMaxMonthHeight {
+      maximumMonthHeightDidChange?(maxMonthHeight)
+    }
+    previousMaxMonthHeight = maxMonthHeight
```
bryankeller, line 244: we're in `layoutSubviews` here fyi

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
```diff
@@ -57,18 +57,17 @@ class DemoViewController: UIViewController {
...
           lessThanOrEqualTo: view.layoutMarginsGuide.trailingAnchor),
         calendarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
-        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375)
+        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375),
+        calendarView.widthAnchor.constraint(equalToConstant: 375).prioritize(at: .defaultLow),
       ])
     case .horizontal:
       NSLayoutConstraint.activate([
         calendarView.centerYAnchor.constraint(equalTo: view.layoutMarginsGuide.centerYAnchor),
-        calendarView.heightAnchor.constraint(equalToConstant: 275),
```
bryankeller, line 65: now the calendar has an intrinsic height, so no height constraint is needed! 😄

Sources/Internal/DoubleLayoutPassHelpers.swift
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  public override func setContentCompressionResistancePriority(
+    _ priority: UILayoutPriority,
+    for axis: NSLayoutConstraint.Axis)
+  {
+    doubleLayoutPassSizingLabel.setContentCompressionResistancePriority(priority, for: axis)
+  }
+
+}
```
bryankeller, line 78: Wanted to keep this [unfortunately necessary] garbage out of `CalendarView.swift`. This extension seemed reasonably well-contained.

Sources/Internal/DoubleLayoutPassHelpers.swift
```diff
@@ -0,0 +1,136 @@
...
+
+  // MARK: Lifecycle
+
+  init(provider: WidthDependentIntrinsicContentHeightProviding) {
+    self.provider = provider
+
+    super.init(frame: .zero)
+
+    numberOfLines = 0
+    isUserInteractionEnabled = false
```
bryankeller, line 100: I know this is the default but just being I'd prefer to be explicit when doing crazy stuff lol

Sources/Internal/DoubleLayoutPassHelpers.swift
```diff
@@ -0,0 +1,136 @@
...
+  override func drawText(in _: CGRect) {
+    // Do nothing, this view doesn't render any visual content.
+  }
+
+  // MARK: Private
+
+  private weak var provider: WidthDependentIntrinsicContentHeightProviding?
+
+}
+
```
bryankeller, line 136: this whole file is just an adaptation of the one I created for the Airbnb app. I removed Airbnb-specific language and made this _less_ general since it's only used internally by `CalendarView`.

Sources/Public/CalendarView.swift
```diff
@@ -972,6 +980,27 @@ extension CalendarView: UIScrollViewDelegate {
 
 }
 
+// MARK: WidthDependentIntrinsicContentHeightProviding
+
+extension CalendarView: WidthDependentIntrinsicContentHeightProviding {
+
+  // This is where we perform our width-dependent height calculation. See `DoubleLayoutPassHelpers`
+  // for more details about why this is needed and how it works.
```
bryankeller, line 988: sorry tim apple

bryankeller (commented):
@bryankeller I've gone with the double-layout-pass approach that we use internally at Airbnb - I'm liking this a lot better. Everything "just works" 😄

bryankeller:
Also - see the updated screenshots / solution description in the PR description!

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
```diff
@@ -57,18 +57,17 @@ class DemoViewController: UIViewController {
...
-        calendarView.leadingAnchor.constraint(
-          greaterThanOrEqualTo: view.leadingAnchor),
-        calendarView.trailingAnchor.constraint(
-          lessThanOrEqualTo: view.trailingAnchor),
+        calendarView.leadingAnchor.constraint(greaterThanOrEqualTo: view.leadingAnchor),
+        calendarView.trailingAnchor.constraint(lessThanOrEqualTo: view.trailingAnchor),
         calendarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
-        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375)
+        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375),
+        calendarView.widthAnchor.constraint(equalToConstant: 375).prioritize(at: .defaultLow),
```
bryankeller, line 70: had to add this since `CalendarView` does _not_ have an intrinsic width, so the horizontal width is ambiguous without it.

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/DemoViewController.swift
```diff
@@ -57,18 +57,17 @@ class DemoViewController: UIViewController {
         calendarView.trailingAnchor.constraint(
           lessThanOrEqualTo: view.layoutMarginsGuide.trailingAnchor),
         calendarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
-        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375)
+        calendarView.widthAnchor.constraint(lessThanOrEqualToConstant: 375),
+        calendarView.widthAnchor.constraint(equalToConstant: 375).prioritize(at: .defaultLow),
```
bryankeller, line 61: had to add this since `CalendarView` does _not_ have an intrinsic width, so the horizontal width is ambiguous without it.

Sources/Public/CalendarView.swift
```diff
@@ -972,6 +980,27 @@ extension CalendarView: UIScrollViewDelegate {
...
+    // Force layout so that `visibleItemsDetails.maxMonthHeight` is calculated.
+    setNeedsLayout()
+    layoutIfNeeded()
+
+    guard let visibleItemsDetails = visibleItemsDetails else {
+      return CGSize(width: UIView.noIntrinsicMetric, height: UIView.noIntrinsicMetric)
+    }
+
+    let height = visibleItemsDetails.maxMonthHeight + visibleItemsDetails.heightOfPinnedContent
+    return CGSize(width: UIView.noIntrinsicMetric, height: height)
```
bryankeller, line 999: note: no intrinsic width. `CalendarView` can resize freely in the horizontal direction, and the only thing that will change is the max month height (and therefore, the intrinsic content height) in response.

Resizing the calendar horizontally does not impact the width - for horizontal calendars, you'll either have extra space at the bottom or you'll clip some content, and for vertical calendars you'll just create a larger or smaller scrollable region by changing the height.

Sources/Internal/DoubleLayoutPassHelpers.swift
```diff
@@ -0,0 +1,136 @@
...
+        "The sizing label's `provider` should not be `nil` for the duration of the its life")
+    }
+    if preferredMaxLayoutWidth == 0 {
+      return super.intrinsicContentSize
+    } else {
+      return provider.intrinsicContentSize(forHorizontallyInsetWidth: preferredMaxLayoutWidth)
+    }
+  }
+
+  override func draw(_: CGRect) {
```
brynbodayle, line 123: I probably mentioned this when implementing in the Airbnb app, but are we sure we need these empty overrides? There's the possibility that they impact performance https://stackoverflow.com/questions/18748276/why-an-empty-implementation-of-drawrect-will-adversely-affect-performance-durin

Perhaps we could just keep the label hidden.

brynbodayle (approved):
One comment on performance, but overall LGTM. Glad we went with this approach despite the magic, should make things easier for folks.

Sources/Internal/DoubleLayoutPassHelpers.swift
bryankeller: Ah yeah that's simpler. If people really go digging through the hierarchy to grab this label and set it visible, that's not really our problem 😛

PR #171, opened 2022-07-27
Add support for custom day shape

markst:
## Details

Adding support for other day view shapes besides circle & rectangle.

## Motivation and Context

We wish to use diamond shape day selector. Rather than adding specific shape needs, this allows defining a closure for any `UIBezierPath` for a given bounds. 

## How Has This Been Tested

Tested in iOS simulator.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/Item Views/Shape.swift
+++ b/Sources/Public/Item Views/Shape.swift
@@ -14,9 +14,39 @@
 // limitations under the License.
 
 import CoreGraphics
+import UIKit.UIBezierPath
 
 /// Represents a background or highlight layer shape; used by `DayView` and `DayOfWeekView`.
-public enum Shape: Hashable {
+public enum Shape {
   case circle
   case rectangle(cornerRadius: CGFloat = 0)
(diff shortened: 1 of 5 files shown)


Sources/Public/Item Views/Shape.swift
```diff
@@ -14,9 +14,39 @@
...
+
+extension Shape: Hashable {
+  public func hash(into hasher: inout Hasher) {
+    switch self {
+    case .circle:
+      break
+    case .rectangle(let cornerRadius):
+      hasher.combine(cornerRadius)
+    case .custom:
+      break
```
markst, line 34: not sure if I should be adding any kind of hash based on closure

bryankeller: Yeah, this will cause issues with view reuse since closures can't be compared for equality.

bryankeller (commented):
Rather than modify the base view, I would recommend just forking the component in your app. Feel free to copy and paste the code in `DayView` and make it work for your specific app.

PR #172, opened 2022-08-17
Support the system scroll-to-top gesture

bryankeller:
## Details

This PR adds support for the system scroll-to-top gesture. Since `HorizonCalendar` handles programmatic scrolling in a custom way, we need to intercept the scroll-to-top request and perform the programmatic scroll ourselves.

## Related Issue

N/A

## Motivation and Context

Airbnb app need

## How Has This Been Tested

Demo app

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -422,7 +422,6 @@ public final class CalendarView: UIView {
 
   private lazy var scrollView: NoContentInsetAdjustmentScrollView = {
     let scrollView = NoContentInsetAdjustmentScrollView()
-    scrollView.scrollsToTop = false
     scrollView.showsVerticalScrollIndicator = false
     scrollView.showsHorizontalScrollIndicator = false
     scrollView.delegate = self
@@ -762,7 +761,7 @@ public final class CalendarView: UIView {
     case (.before, .after), (.after, .before):
       finalizeScrollingTowardItem(for: scrollToItemContext)
 
-      // Force layout immdiately to prevent the overshoot from being visible to the user.
+      // Force layout immediately to prevent the overshoot from being visible to the user.
       setNeedsLayout()
       layoutIfNeeded()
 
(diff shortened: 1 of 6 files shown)


Sources/Public/CalendarView.swift
```diff
@@ -990,6 +997,15 @@ extension CalendarView: UIScrollViewDelegate {
     }
   }
 
+  public func scrollViewShouldScrollToTop(_ scrollView: UIScrollView) -> Bool {
+    if content.monthsLayout.isHorizontal {
+      return false
+    } else {
+      self.scrollView.isAboutToScrollToTopFromSystemGesture = true
+      return true
```
brynbodayle, line 1005: Instead would it be more direct to return false here and trigger the programatic scrolling right here?

bryankeller: That was my initial thinking (and it would definitely make things a bit simpler), but I'm worried that this function could be called / checked at other times by the system. This really should be a side-effect-free way for the system to check whether a scroll view supports the scroll-to-top gesture, and I have no idea if this can be called in other situations no or in the future. Thinking the current approach is a bit more resilient.

bryankeller: Although I guess you could argue that if this was called in a situation where we weren't about to scroll to the top, then the next call to `scrollView.contentOffset`'s setter would also be incorrect... that wouldn't have too bad of an affect - we'd just ignore the first call to the setter. Maybe the contentOffset setter override should also check the target offset to make sure it actually is the top of the scroll view.

bryankeller: So the logic would then be:

1. `scrollViewShouldScrollToTop` called, set `isAboutToScrollToTopFromSystemGesture` to `true`
2. `scrollView.contentOffset.set` called, cancel the scroll only if the target offset y is actually at the top
3. set `isAboutToScrollToTopFromSystemGesture` to `false`

brynbodayle: To me, assuming that the scroll to top tap triggers the scroll view's `contentOffset` setter seems less reliable than assuming that there is a 1:1 relationship between `scrollViewShouldScrollToTop` getting called and the scroll to top gesture. I think returning false from the `scrollViewShouldScrollToTop` function and triggering your custom action is the typical way of customizing this behavior despite the naming which suggests it shouldn't have side-effects. I've used this approach before. And if it's common, Apple would be less likely to change the behavior of it down the road breaking a lot of apps.

bryankeller: I see what you're saying - either way we need to make an assumption somewhere about how Apple is calling into these things. My initial concern was that someone else could actually call this property / function from their app, but since our scroll view is a private implementation detail of the calendar, I don't think we actually need to worry about this.

Latest commit has some changes!

brynbodayle (approved):

PR #173, opened 2022-08-22
Fix navigation bar transparency

bryankeller:
## Details

This fixes navigation bar transparency (when scrolling). This regression was introduced when I added a double-layout-pass sizing `UILabel` to make working with horizontal calendars easier (in https://github.com/airbnb/HorizonCalendar/pull/150). UIKit seems to only observe the scroll position of the first subview of a view, so if the label is subview 0 and the scroll view is subview 1, then the navigation bar doesn't adjust its transparency on scroll. This wasn't an issue on iOS 14, but iOS 15 seems to change the nav bar style a bit.

| Before | After |
| ---- | ---- |
|  Simulator Screen Shot - iPhone 13 Pro - 2022-08-22 at 12 51 27| Simulator Screen Shot - iPhone 13 Pro - 2022-08-22 at 12 51 02 | 

## Related Issue

N/A

## Motivation and Context

A bug I noticed in the example app.

## How Has This Been Tested

Example app

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -57,8 +57,11 @@ public final class CalendarView: UIView {
       backgroundColor = .white
     }
 
-    installDoubleLayoutPassSizingLabel()
+    // Must be the first subview so that `UINavigationController` can monitor its scroll position
+    // and make navigation bars opaque on scroll.
     addSubview(scrollView)
+
+    installDoubleLayoutPassSizingLabel()
 
     setContent(initialContent)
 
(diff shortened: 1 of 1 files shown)


Sources/Public/CalendarView.swift
```diff
@@ -57,9 +57,12 @@ public final class CalendarView: UIView {
       backgroundColor = .white
     }
 
-    installDoubleLayoutPassSizingLabel()
+    // Must be the first subview so that `UINavigationController` can monitor its scroll position
+    // and make navigation bars opaque on scroll.
     addSubview(scrollView)
```
brynbodayle, line 62:
```suggestion
    insertSubview(scrollView, at: 0)
```
makes this code a bit less order dependent and explicit?

brynbodayle (approved):

Sources/Public/CalendarView.swift
bryankeller: Oh yes, great suggestion!

PR #179, opened 2022-10-05
Support animated setContent updates

bryankeller:
## Details

This PR enables us to animate some content update changes by calling `layoutIfNeeded` in a `UIView` animation block. For example, a vertical spacing between weeks can now be animated. We won't animate views that are moving due to view reuse. We also avoid animating views that are being added to the view hierarchy for the first time (existing behavior).

## Related Issue

N/A

## Motivation and Context

Better animation support.

## How Has This Been Tested

Example app in simulator.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -627,14 +627,14 @@ public final class CalendarView: UIView {
 
     reuseManager.viewsForVisibleItems(
       visibleItems,
-      viewHandler: { view, visibleItem, previousBackingVisibleItem in
-        if view.superview == nil {
-          UIView.performWithoutAnimation {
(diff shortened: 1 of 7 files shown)


Sources/Public/CalendarView.swift
```diff
@@ -652,16 +652,13 @@ public final class CalendarView: UIView {
   private func configureView(_ view: ItemView, with visibleItem: VisibleCalendarItem) {
     view.calendarItemModel = visibleItem.calendarItemModel
 
-    // Update the visibility
-    UIView.performWithoutAnimation {
```
bryankeller, line 656: previously, we were preventing animations for frame updates. We only need to prevent animations if an item is being reused. There are legit use cases where the frame might update (and we'd want to animate), like changing the day aspect ratio, or the spacing between elements.

brynbodayle (approved):

PR #182, opened 2022-10-14
Bk/swift UI item model support

bryankeller:
## Details

This adds support for using SwiftUI views with the various `CalendarViewContent` item-provider functions (`dayItemProvider`, `dayRangeItemProvider`, `monthHeaderItemProvider`, etc).

Most of the support to make this possible is done in `SwiftUIWrapperView` - a `UIView` subclass that manages the hosting controller that owns the SwiftUI view.

Here's a comparison of using a simple label as a day view in both UIKit and SwiftUI:

### UIKit

```swift
struct DayLabel: CalendarItemViewRepresentable {

  struct InvariantViewProperties: Hashable {
    let font: UIFont
    let textColor: UIColor
  }

  struct ViewModel: Equatable {
    let day: Day
  }

  static func makeView(
    withInvariantViewProperties invariantViewProperties: InvariantViewProperties)
    -> UILabel
  {
    let label = UILabel()

    label.backgroundColor = .clear
    label.font = invariantViewProperties.font
    label.textColor = invariantViewProperties.textColor
    label.textAlignment = .center
    
    return label
  }

  static func setViewModel(_ viewModel: ViewModel, on view: UILabel) {
    view.text = "\(viewModel.day.day)"
  }

}
```

```swift
  return CalendarViewContent(...)

    .dayItemProvider { day in
      DayLabel.calendarItemModel(
        invariantViewProperties: .init(font: UIFont.systemFont(ofSize: 18),  textColor: .darkGray),
        viewModel: .init(day: day))
    }
```

### SwiftUI

```swift
  return CalendarViewContent(...)

    .dayItemProvider { day in
      Text("\(day.day)")
        .font(.system(size: 18))
        .foregroundColor(Color(UIColor.darkGray))
        .calendarItemModel
    }
```

## Related Issue

N/A

## Motivation and Context

Enable developers to use SwiftUI views with HorizonCalendar.

## How Has This Been Tested

Example app.

## Types of changes

## Checklist

### DIFF
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIItemModelsDemoViewController.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIItemModelsDemoViewController.swift
@@ -1,5 +1,5 @@
-// Created by Bryan Keller on 6/18/20.
-// Copyright © 2020 Airbnb Inc. All rights reserved.
+// Created by Bryan Keller on 8/23/22.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
 
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,15 +15,16 @@
 
 import HorizonCalendar
 import UIKit
+import SwiftUI
 
-final class SelectedDayTooltipDemoViewController: DemoViewController {
+final class SwiftUIItemModelsDemoViewController: DemoViewController {
 
   // MARK: Internal
 
   override func viewDidLoad() {
     super.viewDidLoad()
 
-    title = "Selected Day Tooltip"
+    title = "SwiftUI Day and Month Views"
 
     calendarView.daySelectionHandler = { [weak self] day in
       guard let self = self else { return }
@@ -48,47 +49,47 @@ final class SelectedDayTooltipDemoViewController: DemoViewController {
 
     let selectedDate = self.selectedDate
 
-    let overlaidItemLocations: Set<CalendarViewContent.OverlaidItemLocation>
-    if let selectedDate = selectedDate {
-      overlaidItemLocations = [.day(containingDate: selectedDate)]
-    } else {
-      overlaidItemLocations = []
-    }
-
     return CalendarViewContent(
       calendar: calendar,
       visibleDateRange: startDate...endDate,
       monthsLayout: monthsLayout)
 
       .interMonthSpacing(24)
+      .verticalDayMargin(8)
+      .horizontalDayMargin(8)
 
-      .dayItemProvider { [calendar, dayDateFormatter] day in
-        var invariantViewProperties = DayView.InvariantViewProperties.baseInteractive
-
-        let date = calendar.date(from: day.components)
-        if date == selectedDate {
-          invariantViewProperties.backgroundShapeDrawingConfig.borderColor = .blue
-          invariantViewProperties.backgroundShapeDrawingConfig.fillColor = .blue.withAlphaComponent(0.15)
+      .monthHeaderItemProvider { [calendar, monthDateFormatter] month in
+        guard let firstDateInMonth = calendar.date(from: month.components) else {
+          preconditionFailure("Could not find a date corresponding to the month \(month).")
         }
-
-        return DayView.calendarItemModel(
-          invariantViewProperties: invariantViewProperties,
-          viewModel: .init(
-            dayText: "\(day.day)",
-            accessibilityLabel: date.map { dayDateFormatter.string(from: $0) },
-            accessibilityHint: nil))
+        let monthText = monthDateFormatter.string(from: firstDateInMonth)
+        return HStack {
+          Text(monthText).font(.headline)
+          Spacer()
+        }
+        .padding(.vertical)
+        .calendarItemModel
       }
 
-      .overlayItemProvider(for: overlaidItemLocations) { overlayLayoutContext in
-        TooltipView.calendarItemModel(
-          invariantViewProperties: .init(),
-          viewModel: .init(
-            frameOfTooltippedItem: overlayLayoutContext.overlaidItemFrame,
-            text: "Selected Day"))
+      .dayItemProvider { [calendar, selectedDate] day in
+        let date = calendar.date(from: day.components)
+        let isSelected = date == selectedDate
+        return SwiftUIDayView(dayNumber: day.day, isSelected: isSelected).calendarItemModel
       }
   }
 
   // MARK: Private
+
+  private lazy var monthDateFormatter: DateFormatter = {
+    let dateFormatter = DateFormatter()
+    dateFormatter.calendar = calendar
+    dateFormatter.locale = calendar.locale
+    dateFormatter.dateFormat = DateFormatter.dateFormat(
+      fromTemplate: "MMMM yyyy",
+      options: 0,
+      locale: calendar.locale ?? Locale.current)
+    return dateFormatter
+  }()
 
   private var selectedDate: Date?
 
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/SwiftUIDayView.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/SwiftUIDayView.swift
@@ -1,0 +1,63 @@
+// Created by Bryan Keller on 8/23/22.
+// Copyright © 2022 Airbnb Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import HorizonCalendar
+import SwiftUI
+
+struct SwiftUIDayView: View {
+
+  let dayNumber: Int
+  let isSelected: Bool
+
+  var body: some View {
+    ZStack(alignment: .center) {
+      if #available(iOS 15.0, *) {
+        Circle()
+          .strokeBorder(isSelected ? Color(UIColor.blue) : .clear, lineWidth: 1)
+          .background {
+            Circle()
+              .foregroundColor(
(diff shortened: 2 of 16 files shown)


Example/HorizonCalendarExample/HorizonCalendarExample/SwiftUIDayView.swift
```diff
@@ -0,0 +1,81 @@
...
+          .aspectRatio(1, contentMode: .fill)
+      }
+      Text("\(dayNumber)").foregroundColor(Color.black)
+    }
+    // It's unclear how to make this play nicely with `UIScrollView`'s pan gesture recognizer.
+//    .gesture(
+//      DragGesture(minimumDistance: 0)
+//        .updating($isHighlighted) { _, isHighlighted, _ in
+//          isHighlighted = true
+//        })
```
bryankeller, line 55: Whenever this is recognized, it prevents the scroll view from scrolling the calendar.

Sources/Public/Item Views/DayView.swift
```diff
@@ -302,7 +302,7 @@ extension DayView {
         color = .lightGray
       }
 
-      return DrawingConfig(fillColor: color, borderColor: color)
+      return DrawingConfig(fillColor: color, borderColor: .clear)
```
bryankeller, line 305: Prevents an issue where a transparent color is used, which would make the border visible since it's drawn over the fill color.

Example/HorizonCalendarExample/HorizonCalendarExample/SwiftUIDayView.swift
brynbodayle: What if we wrapped this view in a Button so we can specify a highlight state?

bryankeller: Oh that seems way better - hopefully that plays nicely with the scroll view. Will try it later today.

bryankeller: `Button` doesn't work well since it intercepts taps that would otherwise get passed through to the calendar (which then trigger the `daySelectionHandler` closure). I removed the highlight state code for now, and will explore alternatives later.

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIDayViewDemoViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+    title = "SwiftUI Day View"
+
+    calendarView.daySelectionHandler = { [weak self] day in
+      guard let self = self else { return }
+
+      self.selectedDate = self.calendar.date(from: day.components)
+      self.calendarView.setContent(self.makeContent())
+
+      if UIAccessibility.isVoiceOverRunning, let selectedDate = self.selectedDate {
+        self.calendarView.layoutIfNeeded()
```
erichoracek, line 34: Why do we need this `layoutIfNeeded`? Might be good to have a comment

Example/HorizonCalendarExample/HorizonCalendarExample/SwiftUIDayView.swift
```diff
@@ -0,0 +1,65 @@
...
+            (isSelected ? Color(UIColor.blue.withAlphaComponent(0.15)) : .clear))
+          .aspectRatio(1, contentMode: .fill)
+      }
+      Text("\(dayNumber)").foregroundColor(Color(UIColor.label))
+    }
+  }
+
+}
+
+@available(iOS 13.0.0, *)
```
erichoracek, line 48: I'd assume we can bump the deployment target in the example app above iOS 13 so we don't need these? (even if we keep it lower in the library)

Sources/Public/CalendarView.swift
```diff
@@ -643,9 +645,10 @@ public final class CalendarView: UIView {
         }
       })
 
-    // Remove any old views that weren't reused
+    // Hide any old views that weren't reused. This is cheaper than removing them and re-adding them
+    // as subviews later.
```
erichoracek, line 649: Nice! I'm pretty sure this is what UICollectionView does as well, makes sense

Sources/Public/CalendarItemModel.swift
```diff
@@ -109,3 +110,16 @@ extension CalendarItemViewRepresentable {
...
+// MARK: SwiftUI Convenience Factory
+
+@available(iOS 13.0, *)
+extension View {
+
+  public var calendarItemModel: CalendarItemModel<SwiftUIWrapperView<Self>> {
+    CalendarItemModel<SwiftUIWrapperView<Self>>(
+      invariantViewProperties: .init(initialContent: self),
+      viewModel: .init(content: self))
+  }
```
erichoracek, line 123: Might be useful to add some documentation comments, and detail the fact that this adds a UIHostingController as a child view controller of the closest VC so it doesn't require folks to look through the implementation to find out why the `UIViewController.children` changed

Sources/Public/Item Views/SwiftUIWrapperView.swift
```diff
@@ -0,0 +1,198 @@
...
+
+    init(initialContent: Content) {
+      self.initialContent = initialContent
+    }
+
+    // MARK: Public
+
+    public static func == (lhs: InvariantViewProperties, rhs: InvariantViewProperties) -> Bool {
+      type(of: lhs.initialContent) == type(of: rhs.initialContent)
+    }
```
erichoracek, line 121: Since the `InvariantViewProperties` is nested within the `SwiftUIWrapperView` type which is genercized on `Content` won't this always be true?

Sources/Public/Item Views/SwiftUIWrapperView.swift
```diff
@@ -0,0 +1,198 @@
...
+  /// Recursively traverses up the responder chain to find the closest view controller.
+  fileprivate func closestViewController() -> UIViewController? {
+    self as? UIViewController ?? next?.closestViewController()
+  }
+}
+
+// MARK: - HostingController
+
+@available(iOS 13.0, *)
+private final class HostingController<Content: View>: UIHostingController<Content> {
```
erichoracek, line 176: It might make sense to make this type public or give it a more recognizable name so that it can easily be excluded from UIViewController-based viewDidAppear swizzling impression systems, just in case anyone has something like that

erichoracek (approved):
@bryankeller Looks great! Just a few small suggestions/questions, but the overall approach looks like it will work well 👍

Example/HorizonCalendarExample/HorizonCalendarExample/SwiftUIDayView.swift
bryankeller: That simplifies a few things, thanks for noticing this 😄

Sources/Public/CalendarItemModel.swift
bryankeller: Good call!

Sources/Public/Item Views/SwiftUIWrapperView.swift
bryankeller: Great point, yes!

Sources/Public/Item Views/SwiftUIWrapperView.swift
```diff
@@ -0,0 +1,222 @@
...
+// MARK: - SwiftUIWrapperView
+
+/// A wrapper view that enables SwiftUI `View`s to be used with `CalendarItemModel`s.
+///
+/// Consider using the `calendarItemModel` property, defined as an extension on SwiftUI's`View`, to avoid needing to work with
+/// this wrapper view directly.
+/// e.g. `Text("\(dayNumber)").calendarItemModel`
+///
+/// - Warning: Using a SwiftUI view with the calendar will cause `SwiftUIView.HostingController`(s) to be added to the
+/// closest view controller in the responder chain in relation to the `CalendarView`.
```
bryankeller, line 27: Added a warning here WRT the hosting controller

Sources/Public/CalendarItemModel.swift
```diff
@@ -109,3 +110,23 @@ extension CalendarItemViewRepresentable {
...
+@available(iOS 13.0, *)
+extension View {
+
+  /// Creates a `CalendarItemModel` from a SwiftUI `View`.
+  ///
+  /// This is equivalent to manually creating a
+  /// `CalendarItemModel<SwiftUIWrapperView<YourView>>`, where `YourView` is some SwiftUI `View`.
+  ///
+  /// - Warning: Using a SwiftUI view with the calendar will cause `SwiftUIView.HostingController`(s) to be added to the
+  /// closest view controller in the responder chain in relation to the `CalendarView`.
```
bryankeller, line 125: Added a warning here WRT the hosting controller

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIItemModelsDemoViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+          preconditionFailure("Could not find a date corresponding to the month \(month).")
+        }
+        let monthText = monthDateFormatter.string(from: firstDateInMonth)
+        return HStack {
+          Text(monthText).font(.headline)
+          Spacer()
+        }
+        .padding(.vertical)
+        .calendarItemModel
+      }
```
bryankeller, line 72: added another example - this time of using an inline SwiftUI view for a month header.

Sources/Internal/Calendar+Helpers.swift
```diff
@@ -28,8 +28,7 @@ extension Calendar {
   }
 
   func firstDate(of month: Month) -> Date {
-    let firstDateComponents = DateComponents(era: month.era, year: month.year, month: month.month)
-    guard let firstDate = date(from: firstDateComponents) else {
+    guard let firstDate = date(from: month.components) else {
```
bryankeller, line 31: simplified since there was already an helper to turn a `Month` into `DateComponents`

Sources/Public/Item Views/SwiftUIWrapperView.swift
```diff
@@ -0,0 +1,222 @@
...
+  }
+
+  public override func systemLayoutSizeFitting(
+    _ targetSize: CGSize,
+    withHorizontalFittingPriority horizontalFittingPriority: UILayoutPriority,
+    verticalFittingPriority: UILayoutPriority)
+    -> CGSize
+  {
+    hostingController.sizeThatFits(in: targetSize)
+  }
```
bryankeller, line 68: Had to add this to ensure that month headers self-size correctly.

erichoracek (approved):

PR #184, opened 2022-11-14
Add visibleCentermostMonth in CalendarView

martinpucik:
## Details

- Added a new public getter for month of the most centered calendar layout item

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -120,6 +120,11 @@ public final class CalendarView: UIView {
     visibleItemsDetails?.visibleDayRange
   }
 
+  /// The month of the centered item the layout of the calendar.
+  public var visibleCentermostMonth: Month? {
+    visibleItemsDetails?.centermostLayoutItem.itemType.month
+  }
+
   /// `CalendarView` only supports positive values for `layoutMargins`. Negative values will be changed to `0`.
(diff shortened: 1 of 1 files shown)


bryankeller (changes requested):
Thanks for the contribution @martinpucik ! Can you explain the motivation for this change? What are you trying to do in your app? Why can't you use `visibleMonthRange` or maybe `visibleDayRange`?

I'm not sure the concept of a "visible centermost month" makes a lot of sense. For example, if there are 2 months displayed in the calendar, and they're both fully visible, then which one is the "visible centermost month"? The current implementation relies on an internal layout implementation detail to answer this question, which is probably not what we want.

bryankeller:
@martinpucik the visible month range will include partially visible months. That means that if even one pixel of a month's contents is on screen, it will be part of the `visibleMonthRange`. I could see why this might cause trouble in your case.

What if you instead access the `visibleDayRange`? You can look at its upper and lower bound, then take some date in the middle to approximate what's in the center of the screen. You'll need to do some conversions to `Date`s using your `Calendar` (`NSCalendar`) instance, but it shouldn't be too hard!

martinpucik:
@bryankeller thank you for the feedback! I'll dwell on this a bit more 👍

PR #191, opened 2022-12-15
Bk/fix z index ordering

bryankeller:
## Details

This PR fixes z-axis ordering issues, and re-enables a small performance optimization to hide / unhide subviews, rather than adding and removing subviews when reusing views. I also renamed the internal type `VisibleCalendarItem` -> `VisibleItem`.

Originally, I thought it was fine to just set the `layer.zPosition` to control z-axis ordering, rather than inserting subviews in the correct order in the `subviews` array. Over time, I noticed some issues:

- snapshotting a view would result in the views having the wrong z-axis order in the snapshot, despite things looking fine while using the app
- unhiding a view would result in the view having the wrong z-axis order, despite things looking fine before the view was hidden

I had a workaround for the snapshotting issue, but it was a bit of a hack. I did not have a workaround for the view hiding / unhiding issue, so I figured I should fix the core issue (subviews being inserted in the wrong order into the `subviews` array).

## Related Issue

N/A

## Motivation and Context

- Remove a hack
- Enable a new performance optimization

## How Has This Been Tested

Example app on device, unit tests

## Types of changes

## Checklist


Sources/Internal/RenderInContextObservingLayer.swift
```diff
@@ -1,19 +0,0 @@
-// Created by Bryan Keller on 4/15/21.
-// Copyright © 2021 Airbnb Inc. All rights reserved.
-
-import UIKit
-
-/// A `CALayer` subclass that enables an observer to be notified when `render(in:)` is called (usually for the purpose of
-/// doing some work before a layer is snapshotted). This is needed so that `CalendarView` can sort its
-/// `scrollView.sublayers` array before snapshotting, since `render(in:)` does not respect `CALayer` `zPosition`
-///  values.
-final class RenderInContextObservingLayer: CALayer {
```
bryankeller, line 10: not needed anymore (this was used for the snapshotting workaround)

Sources/Internal/VisibleItem.swift
```diff
@@ -87,18 +87,6 @@ extension VisibleCalendarItem {
...
-      case .layoutItemType: return 500
-      case .pinnedDayOfWeek: return 1000
-      case .pinnedDaysOfWeekRowBackground: return 999
-      case .pinnedDaysOfWeekRowSeparator: return 1001
-      case .daysOfWeekRowSeparator: return 501
-      case .dayRange: return 250
-      case .overlayItem: return 750
-      }
-    }
-
```
bryankeller, line 101: not needed anymore, as this ordering info is encoded in the `SubviewsManager`

Sources/Public/CalendarView.swift
```diff
@@ -47,10 +47,6 @@ public final class CalendarView: UIView {
 
     super.init(frame: .zero)
 
-    (layer as! RenderInContextObservingLayer).willRenderInContext = { [weak self] in
-      self?.sortScrollViewSublayersByZPositions()
-    }
-
```
bryankeller, line 53: no more custom layer needed to run some code before snapshotting a view

Sources/Public/CalendarView.swift
```diff
@@ -78,8 +74,6 @@ public final class CalendarView: UIView {
 
   // MARK: Public
 
-  public override class var layerClass: AnyClass { RenderInContextObservingLayer.self }
-
```
bryankeller, line 82: no more custom layer needed to run some code before snapshotting a view

Sources/Public/CalendarView.swift
```diff
@@ -617,43 +613,43 @@ public final class CalendarView: UIView {
...
         }
       })
 
-    // Remove any old views that weren't reused
-    for visibleItemWithViewToRemove in visibleItemsWithViewsToRemove {
-      previousVisibleViewsForVisibleItems[visibleItemWithViewToRemove]?.removeFromSuperview()
+    // Hide any old views that weren't reused. This is faster than adding / removing subviews.
+    for (_, viewToHide) in viewsToHideForVisibleItems {
+      viewToHide.isHidden = true
     }
```
bryankeller, line 646: new perf optimization to avoid adding / removing subviews more than needed (we can just hide / unhide)

Sources/Public/CalendarView.swift
```diff
@@ -831,12 +827,6 @@ public final class CalendarView: UIView {
...
   }
 
-  /// This is needed to ensure that the `scrollView.sublayers` array is sorted according to each sublayer's `zPosition`.
-  /// This prevents z-index-related rendering issues when a `CalendarView` is being snapshotted via `CALayer.render(in:)`.
-  private func sortScrollViewSublayersByZPositions() {
-    scrollView.layer.sublayers?.sort { $0.zPosition < $1.zPosition }
-  }
-
```
bryankeller, line 839: no more custom layer needed to run some code before snapshotting a view

Sources/Internal/SubviewsManager.swift
```diff
@@ -0,0 +1,105 @@
...
+  // MARK: Lifecycle
+
+  init(parentView: UIView) {
+    self.parentView = parentView
+  }
+
+  // MARK: Internal
+
+  // Only used for XCTest support
+  var _testSupport_insertedItemTypes = [VisibleItem.ItemType]()
```
brynbodayle, line 31: Hmm, would be nice to avoid this. What if this class's `insertSubview` returned the insertion index? Would that be sufficient for testing? Or this class could be simplified to not even manage the view's itself, but just manage indexes.

bryankeller: Good call! Refactored in latest commit

brynbodayle (approved):
Awesome!

PR #205, opened 2023-01-30
Add month grid background view and demo

bryankeller:
## Details

Adds a MonthGridBackgroundView so people can easily add grid lines to their calendar (with the new monthBackgroundItemProvider API).

I'm going to do a follow-up PR with updates to documentation. There are a few things I want to change (including some updated screenshots).

| Grid |
|----|
| Simulator Screen Shot - iPhone 14 Pro - 2023-01-30 at 13 32 10 |

## Related Issue

N/A

## Motivation and Context

Grid are a common use case, so providing a default view seems helpful!

## How Has This Been Tested

Example app

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/ItemViews/MonthGridBackgroundView.swift
+++ b/Sources/Public/ItemViews/MonthGridBackgroundView.swift
@@ -1,0 +1,125 @@
+// Created by Bryan Keller on 1/30/23.
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


Sources/Public/ItemViews/MonthGridBackgroundView.swift
```diff
@@ -0,0 +1,125 @@
...
+
+  // MARK: Public
+
+  public override func draw(_: CGRect) {
+    let context = UIGraphicsGetCurrentContext()
+    context?.setLineWidth(invariantViewProperties.lineWidth)
+    context?.setStrokeColor(invariantViewProperties.color.cgColor)
+
+    if traitCollection.layoutDirection == .rightToLeft {
+      transform = .init(scaleX: -1, y: 1)
```
brynbodayle, line 43: Hmm, I haven't seen this done before I'm not sure if drawRect is a good place to change the transform. Could we just the origin.x of the rects instead? Also, I think we might need to call `setNeedsDisplay` in respond to `traitCollectionDidChange`.

brynbodayle (approved):

Issue #38, opened 2020-08-03
Any plans to support this library on SwiftUI?


sethi-ishmeet:

bryankeller:
Hi  @sethi-ishmeet - this is something that's definitely on our radar. My understanding is that support could be added by wrapping it with `UIViewRepresentable`. I'll be exploring this in the coming weeks!

aakrem:
Hi @bryankeller I've tried wrapping it into a UIViewRepresentable without any success.. 
Any updates on how we can make this library working with swiftui ?

PwBwolf:
This worked for me 

    import SwiftUI
    import HorizonCalendar

    struct HorrizonView: UIViewRepresentable {
   
    
    
    func makeUIView(context: Context) -> CalendarView {
        let calendarView = CalendarView(initialContent: makeContent())
        return calendarView
    }
    
    func updateUIView(_ uiView: CalendarView, context: UIViewRepresentableContext<HorrizonView>) {}
    
    private func makeContent() -> CalendarViewContent {
        let calendar = Calendar(identifier: .gregorian)
        let startDate = calendar.date(from: DateComponents(year: 2020, month: 01, day: 01))!
        let endDate = calendar.date(from: DateComponents(year: 2021, month: 12, day: 31))!
        return CalendarViewContent(
          calendar: calendar,
          visibleDateRange: startDate...endDate,
          monthsLayout: .vertical(options: VerticalMonthsLayoutOptions()))
    }
    
    }

Make sure you call struct in the swiftUI view you wish to display it. 

     struct DatePickerView: View {
    @State private var beginingDate = Date()
    @State private var endDate = Date()
    var body: some View {
        VStack {
            HStack {
                Text(beginingDate, style: .date)
                Image("arrpwshape.zigzag.right.fill")
                Text(endDate, style: .date)
            }
            HorrizonView()

            Button("Save", action: {
                print("Saving dates")
                print($beginingDate)
            })
        }
    }
    
<img width="436" alt="Screen Shot 2020-10-27 at 8 59 20 PM" src="https://user-images.githubusercontent.com/8434000/97377832-4b74b480-1897-11eb-9c62-fdf259f4d62c.png">
}

bryankeller:
@sethi-ishmeet @PwBwolf if you're still actively using this, a pull request to add this SwiftUI support is welcome since it sounds like you've figured out how to do it 🙂 !

fluffy919:
Hi, I have an issue with 
[...]

PR #209, opened 2023-02-07
Bk/calendar view representable

bryankeller:
## Details

This PR introduces `CalendarViewRepresentable` - a way to use `CalendarView` from SwiftUI. I also decided to include the public function docs in this PR. They're identical to the ones used by the `CalendarViewContent` modifiers, with minor wording changes to not actually reference `CalendarViewContent`.

Here's a little video demo of what it's like to use this API:

https://user-images.githubusercontent.com/746571/217365304-8580899c-a3b3-4f91-a1fc-6e36e2175ae0.mov

I still need to add documentation in the README.md. Will do in a future PR before the v2.0.0 release.

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/38

## Motivation and Context

Of course we want great SwiftUI API!

## How Has This Been Tested

Example app

## Types of changes

## Checklist

### DIFF
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIScreenDemoViewController.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIScreenDemoViewController.swift
@@ -99,13 +99,15 @@ struct SwiftUIScreenDemo: View {
       let monthHeaderText = monthDateFormatter.string(from: calendar.date(from: month.components)!)
       if case .vertical = monthsLayout {
         return HStack {
-          Text(monthHeaderText).font(.title2)
+          Text(monthHeaderText)
+            .font(.title2)
           Spacer()
         }
         .padding()
         .calendarItemModel
       } else {
-        return Text(monthHeaderText).font(.title2)
+        return Text(monthHeaderText)
+          .font(.title2)
           .padding()
           .calendarItemModel
       }
--- a/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIScreenDemoViewController.swift
+++ b/Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIScreenDemoViewController.swift
@@ -1,0 +1,176 @@
+// Created by Bryan Keller on 2/1/23.
+// Copyright © 2023 Airbnb Inc. All rights reserved.
+
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import HorizonCalendar
+import SwiftUI
+
+// MARK: - SwiftUIScreenDemoViewController
+
+final class SwiftUIScreenDemoViewController: UIViewController, DemoViewController {
+
+  // MARK: Lifecycle
+
+  init(monthsLayout: MonthsLayout) {
+    self.monthsLayout = monthsLayout
+    super.init(nibName: nil, bundle: nil)
+  }
+
+  required init?(coder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  // MARK: Internal
+
+  let calendar = Calendar.current
+  let monthsLayout: MonthsLayout
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
+
+    title = "SwiftUI Screen"
+
+    let hostingController = UIHostingController(
+      rootView: SwiftUIScreenDemo(calendar: calendar, monthsLayout: monthsLayout))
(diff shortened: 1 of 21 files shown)


Sources/Public/CalendarViewRepresentable.swift
```diff
@@ -0,0 +1,331 @@
...
+  ///   - monthsLayout: The layout of months - either vertically or horizontally.
+  ///   - dataDependency: Any data or state that, when updated, should cause the calendar to re-render. This is needed because
+  ///   `CalendarView` lazily invokes item provider closures outside of the normal SwiftUI update loop. To ensure that the calendar
+  ///   is updated at the right times, pass in any properties that are accessed in your item provider closures. For example, if a
+  ///   `dayItemProvider` closure accesses a local state variable called `selectedDay`, then pass it in here.
+  public init(
+    calendar: Calendar = Calendar.current,
+    visibleDateRange: ClosedRange<Date>,
+    monthsLayout: MonthsLayout,
+    dataDependency: Any? = nil)
```
bryankeller, line 53: Ah, the `dataDependency`... I thought I could avoid it if I just forced layout, which would trigger our item provider closures to get called, but then I realized the problem is unsolvable without eagerly calling every item provider closure with every possible input. Let me try to explain:

SwiftUI calls the `updateUIView` function, and checks to see which state is referenced during that update pass (imagine every `@State` property has a monitor to see when its `get { ... }` is invoked). SwiftUI then creates a list of dependencies for your view, and any time one of those dependencies changes, your view's `updateUIView` will be invoked again.

For `CalendarView`, you might read state inside the `dayItemProvider` closure, for example, to see if the `@State var selectedDay: Day?` is equal to the current day passed into the closure - and if it is, you change the style of the day to appear selected.

But what if you don't always read the `selectedDay` state in your `dayItemProvider` closure? Maybe your `dayItemProvider` has some code that looks like this:

```swift
.dayItemProvider { day in 
  if day.isCurrentYear {
    return DefaultDayView(day).calendarItemModel
  } else {
    let isSelected = day == selectedDay // The `selectedDay` state is only read here!
    return CustomDayView(day, isSelected: isSelected).calendarItemModel
  }
}
```

When the `CalendarView` is first displayed, it's only going to invoke the `dayItemProvider` for the currently-visible days (likely just 1 or 2 months worth of days). This means that the first time `updateUIView` is called, SwiftUI is going to see that `selectedDay` was never read and is therefore _not_ a dependency for our view. Later on when `selectedDay` changes, `updateUIView` will not be called because `selectedDay` was already determined to not be a dependency.

I don't see a way around this. Maybe there's some kind of hack where we can trigger view updates manually? We still have time to figure out an alternative before I do an official v2.0.0 tag.

Example/HorizonCalendarExample/HorizonCalendarExample/SwiftUIDayView.swift
```diff
@@ -23,21 +23,13 @@ struct SwiftUIDayView: View {
...
-            (isSelected ? Color(UIColor.blue.withAlphaComponent(0.15)) : .clear))
-          .aspectRatio(1, contentMode: .fill)
-      }
+      Circle()
+        .strokeBorder(isSelected ? Color.accentColor : .clear, lineWidth: 2)
+        .background {
+          Circle()
+            .foregroundColor(isSelected ? .white : .clear)
+        }
+        .aspectRatio(1, contentMode: .fill)
```
bryankeller, line 32: bumped the base SDK to 16+ in the example project to clean this up

Sources/Public/CalendarViewRepresentable.swift
```diff
@@ -0,0 +1,331 @@
...
+// MARK: Content Modifiers
+
+@available(iOS 13.0, *)
+extension CalendarViewRepresentable {
+
+  public func dayAspectRatio(_ dayAspectRatio: CGFloat) -> Self {
+    var view = self
+    view.dayAspectRatio = dayAspectRatio
+    return view
+  }
```
bryankeller, line 182: all these need documentation. will do in the next PR.

Example/HorizonCalendarExample/HorizonCalendarExample/Demo View Controllers/SwiftUIScreenDemoViewController.swift
```diff
@@ -0,0 +1,176 @@
...
+
+    .verticalDayMargin(8)
+    .horizontalDayMargin(8)
+    .interMonthSpacing(16)
+
+    .monthHeaderItemProvider { month in
+      let monthHeaderText = monthDateFormatter.string(from: calendar.date(from: month.components)!)
+      if case .vertical = monthsLayout {
+        return HStack {
+          Text(monthHeaderText).font(.title2)
```
brynbodayle, line 102:
```suggestion
          Text(monthHeaderText)
          .font(.title2)
```

move these to the next line for consistency?

Sources/Public/CalendarViewRepresentable.swift
brynbodayle: Hmm, this is a tricky one. At the highest level it seems like the problem is trying to bridge a lazy UIKit-based view to SwiftUI. 

I think the current solution is okay though given the constraints and no other viable options at the moment. Documentation is clear. What if we removed the default value for `dataDependency` to encourage engineers to think about this and avoid potential bugs?

bryankeller: > At the highest level it seems like the problem is trying to bridge a lazy UIKit-based view to SwiftUI.
Yep

Good call on removing the default, I agree

PR #212, opened 2023-02-16
ml/Fix no ended state fires for multipleDaySelectionDragHandler

xingsuo:
## Details

This change fixed the issue that multipleDaySelectionDragHandler never get called when the gestureRecognizer is in `ended` state. 

## Related Issue

N/A

## Motivation and Context

We need this state to fire in order to let the user know the dragging gesture has been ended.

## How Has This Been Tested

Tested in the app which has been integrated with HorizonCalendar.

## Types of changes

## Checklist

- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -510,7 +510,7 @@ public final class CalendarView: UIView {
 
   private lazy var scrollViewDelegate = ScrollViewDelegate(calendarView: self)
   private lazy var gestureRecognizerDelegate = GestureRecognizerDelegate(calendarView: self)
-
+  
   // Necessary to work around a `UIScrollView` behavior difference on Mac. See `scrollViewDidScroll`
   // and `preventLargeOverScrollIfNeeded` for more context.
   private lazy var isRunningOnMac: Bool = {
@@ -519,7 +519,7 @@ public final class CalendarView: UIView {
         return true
       }
     }
-  
+    
     return false
   }()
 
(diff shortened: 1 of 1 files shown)


Sources/Public/CalendarView.swift
```diff
@@ -939,16 +939,16 @@ public final class CalendarView: UIView {
     let newMultipleDaySelectionDay: Day
```
bryankeller, line 939:
```suggestion
```

Sources/Public/CalendarView.swift
```diff
@@ -939,16 +939,16 @@ public final class CalendarView: UIView {
...
     if let intersectedDay, intersectedDay != lastMultipleDaySelectionDay {
       newMultipleDaySelectionDay = intersectedDay
```
bryankeller, line 941:
```suggestion
```

Sources/Public/CalendarView.swift
```diff
@@ -939,16 +939,16 @@ public final class CalendarView: UIView {
...
-    } else {
-      return
-    }
 
-    lastMultipleDaySelectionDay = newMultipleDaySelectionDay
-
-    multipleDaySelectionDragHandler?(newMultipleDaySelectionDay, gestureRecognizer.state)
+      lastMultipleDaySelectionDay = newMultipleDaySelectionDay
+      multipleDaySelectionDragHandler?(newMultipleDaySelectionDay, gestureRecognizer.state)
```
bryankeller, line 944:
```suggestion
      lastMultipleDaySelectionDay = intersectedDay
      multipleDaySelectionDragHandler?(intersectedDay, gestureRecognizer.state)
```

bryankeller (changes requested):
Thanks for finding this bug and fixing it! I have a few suggestions to simplify the logic. Let me know if they make sense.

xingsuo:
> Thanks for finding this bug and fixing it! I have a few suggestions to simplify the logic. Let me know if they make sense.

@bryankeller updated!

bryankeller (approved):

PR #217, opened 2023-03-23
Bk/auto scroll drag gesture

bryankeller:
## Details

This adds auto-scroll functionality to the multi-day-selection drag gesture. If a user starts a drag gesture, then drags close to the edge of the scrollable content, the calendar will auto-scroll forward or backward in time so that they can reach dates outside of the original viewport.

## Related Issue

N/A

## Motivation and Context

Airbnb app need

## How Has This Been Tested

Tested in simulator and on real iPhone, vertical and horizontal layout.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Internal/ScrollMetricsMutator.swift
+++ b/Sources/Internal/ScrollMetricsMutator.swift
@@ -76,7 +76,10 @@ final class ScrollMetricsMutator {
 
   func applyOffset(_ offset: CGFloat) {
     let currentOffset = scrollMetricsProvider.offset(for: scrollAxis)
-    scrollMetricsProvider.setOffset(to: currentOffset + offset, for: scrollAxis)
+    let minimumOffset = scrollMetricsProvider.minimumOffset(for: scrollAxis)
+    let maximumOffset = scrollMetricsProvider.maximumOffset(for: scrollAxis)
+    let newOffset = max(minimumOffset, min(currentOffset + offset, maximumOffset))
+    scrollMetricsProvider.setOffset(to: newOffset, for: scrollAxis)
   }
 
   // MARK: Private
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -502,6 +502,9 @@ public final class CalendarView: UIView {
   private weak var scrollToItemDisplayLink: CADisplayLink?
   private var scrollToItemAnimationStartTime: CFTimeInterval?
 
+  private weak var autoScrollDisplayLink: CADisplayLink?
+  private var autoScrollOffset: CGFloat?
+
   private var cachedAccessibilityElements: [Any]?
   private var focusedAccessibilityElement: Any?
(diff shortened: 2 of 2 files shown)


Sources/Internal/ScrollMetricsMutator.swift
```diff
@@ -76,7 +76,10 @@ final class ScrollMetricsMutator {
 
   func applyOffset(_ offset: CGFloat) {
     let currentOffset = scrollMetricsProvider.offset(for: scrollAxis)
-    scrollMetricsProvider.setOffset(to: currentOffset + offset, for: scrollAxis)
+    let minimumOffset = scrollMetricsProvider.minimumOffset(for: scrollAxis)
+    let maximumOffset = scrollMetricsProvider.maximumOffset(for: scrollAxis)
+    let newOffset = max(minimumOffset, min(currentOffset + offset, maximumOffset))
+    scrollMetricsProvider.setOffset(to: newOffset, for: scrollAxis)
```
bryankeller, line 82: this caps the scroll offset to be within the valid range of offset values

Sources/Public/CalendarView.swift
```diff
@@ -162,6 +162,14 @@ public final class CalendarView: UIView {
     setNeedsLayout()
   }
 
+  public override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+    // If the user interacts with the calendar, we should clear out any existing
+    // `scrollToItemContext` that might be leftover from the initial layout process.
+    scrollToItemContext = nil
```
bryankeller, line 168: I used to do this in `scrollViewDidScroll`, but that case didn't cover the drag gesture auto scroll that this PR adds. The real logic is "whenever the user interacts with the calendar, set this to nil" - I think `hitTest` is the most reliable place? `touchesBegan` doesn't get called if a different view or gesture captures the touch.

brynbodayle: hmmm, I think this function having side effects isn't ideal, maybe is risky.

> leftover from the initial layout process.

 can you elaborate? Hopefully we can find a better place to put this

bryankeller: We use a `scrollToItemContext` to represent a programmatic scroll to a certain item / position (with or without animation).

When the calendar's bounds changes, we set a non-animated `scrollToItemContext` to scroll to the same offset that we were at prior to the bounds change. This enables us to maintain our scroll position on bounds changes (rotation, initial layout as the calendar receives its initial/final size).

For as long as there's a non-nil `scrollToItemContext`, the calendar won't scroll like normal. That's why the previous logic would set it to nil in `scrollViewDidScroll`, effectively enabling the calendar to scroll normally. We need to do this same logic for the auto-scroll behavior that this PR adds.

I could just duplicate the logic to nil out the `scrollToItemContext` in the auto-scroll code. I thought maybe I could solve it a little bit earlier by detecting any kind of user interaction.

bryankeller:
Pushed some updates based on review feedback @brynbodayle

brynbodayle (approved):

PR #249, opened 2023-07-25
Fixes a misbehaviour of dayOfWeekAspectRatio modificator in SwiftUI r…

hxml16:
dayOfWeekAspectRatio was always nil and had no effect


bryankeller (approved):
Thank you!

PR #262, opened 2023-08-30
Fix implicit animationg bug

bryankeller:
## Details

This fixes an issue that could cause the calendar to layout with an odd animation. Recently, we changed the way `layoutSubviews` works - we check if we're currently in an animation closure, and if we are, we layout a larger region than what's currently visible so that offscreen items animate in / away correctly. 

This behavior change only makes sense if the reason we're in `layoutSubviews` is because of a call to `setContent` (which will do some animation preparation). If we end up in `layoutSubviews` in an animation closure without `setContent` getting called first, then we didn't do any animation preparation in `setContent`, and the new `layoutSubviews` behavior doesn't make sense anymore.

The solution is to set a flag in `setContent` to indicate that the next call to `layoutSubviews` should use the new behavior for animated updates, rather than checking if we're in an animation closure (which could happen if a random part of code calls `layoutIfNeeded` in an animation closure without also calling `setContent` first).

## Related Issue

N/A

## Motivation and Context

Found a bug in the Airbnb app

## How Has This Been Tested

Example app, Airbnb app

## Types of changes

## Checklist


brynbodayle (dismissed):
Makes sense!

bryankeller:
@brynbodayle I fixed another (related) issue with pinned days of the week. I was drawing them at weird positions due to the whole 3x visible bounds thing that I added. The 2 new commits fix that + add unit tests for it.

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -242,7 +248,7 @@ final class VisibleItemsProvider {
 
     // Handle pinned day-of-week layout items
     if case .vertical(let options) = content.monthsLayout, options.pinDaysOfWeekToTop {
-      handlePinnedDaysOfWeekIfNeeded(yContentOffset: bounds.minY, context: &context)
+      handlePinnedDaysOfWeekIfNeeded(yContentOffset: offset.y, context: &context)
```
bryankeller, line 251: we don't want to use the extended bounds for this, otherwise we'll render these pinned items off screen

jpsoultanis (approved):

brynbodayle (approved):

Issue #256, opened 2023-08-22
swiftUI add StateObject to SwiftUIDayView


buzai:
I add a @StateObject to the SwiftUIDayView, I want get some data by date and then show some tags in the SwiftUIDayView view,
The first launch seemed normal, but I found that once I swiped month, the data was wrong, and it seemed that the SwiftUIDayView task only  exec once?

bryankeller:
Could you provide some sample code, or a minimal-repro case?

buzai:
@bryankeller 

```

import SwiftUI
import HorizonCalendar
@MainActor
class RecordForDateModel: ObservableObject {
    @Published var records: String = ""
    func fetchRecordsWithDate(year: Int, month: Int, day: Int) async {
        print("fetchRecordsWithDate")
        records = "\(day)"
    }
}

struct SwiftUIDayView: View {

    let dayNumber: Int
    let yearNumber: Int
    let monthNumber: Int

    let isSelected: Bool
    @StateObject var recordItemModel = RecordForDateModel()
    
    var body: some View {
        ZStack(alignment: .center) {
            Circle()
                .strokeBorder(isSelected ? Color.accentColor : .clear, lineWidth: 2)
                .background {
                    Circle()
                        .foregroundColor(isSelected ? Color(UIColor.systemBackground) : .clear)
                }
                .aspectRatio(1, contentMode: .fill)
            VStack {
                Text("\(dayNumber)").foregroundColor(Color(UIColor.label))
                Text("\(recordItemModel.records.description)").foregroundColor(Color(UIColor.label))
            }
            

        }.task {
            await recordItemModel.fetchRecordsWithDate(year: yearNumber, month: monthNumber, day: dayNumber)
        }
    }

}


// MARK: - SwiftUIScreenDemo

struct SwiftUIScreenDemo: View {

  // MARK: Lifecycle

  init(calendar: Calendar, monthsLayout: MonthsLayout) {
    self.calendar = calendar
    self.monthsLayout = monthsLayout

    let startDate = calendar.date(from: DateComponents(year: 2023, month: 01, day: 01))!
    let endDate = calendar.date(from: DateComponents(year: 2026, month: 12, day: 31))!
    visibleDateRange = startDate...endDate

    monthDateFormatter = DateFormatter()
    monthDateFormatter.calendar = calendar
    monthDateFormatter.locale = calendar.locale
    monthDateFormatter.dateFormat = DateFormatter.dateFormat(
[...]

PR #265, opened 2023-09-08
Fix SwiftUI item view onAppear not getting called

bryankeller:
## Details

This fixes onAppear not getting called when SwiftUI item views were reused. The core issue is that, due to views being reused under the hood, updating the `rootView` of a hosting controller from `ReusedHostingController_ofViewType_DayView(3)` to `ReusedHostingController_ofViewType_DayView(19)` would only trigger a content / data update (the text changing from `3` to `19`), _not an identity update_, which is what's really happening here. View reuse is an implementation detail, but it is critical for performance.

To demonstrate the issue, I've made a simple SwiftUI day view that reproduces the problem. Here, we update a a local store in `onAppear`, and use that value as a text overlay, making it easy to compare that value vs. the expected value. If there's a reuse issue, the numbers don't match, and we make the overlaid text red.

| Reuse issue | After fix |
| ---- | ---- |
| Simulator Screen Recording - iPhone 14 Pro - 2023-09-08 at 11 26 51 | Simulator Screen Recording - iPhone 14 Pro - 2023-09-08 at 11 26 12 |

```swift
private final class Store: ObservableObject {
  @Published var value: Int?
}

struct SwiftUIDayView: View {

  let dayNumber: Int

  @StateObject private var store = Store()

  var body: some View {
    ZStack(alignment: .center) {
      Text("\(dayNumber)")

      if let storeValue = store.value {
        Text("\(storeValue)")
          .foregroundColor((storeValue == dayNumber ? Color.green : Color.red).opacity(0.5))
      }
    }
    .onAppear {
      store.value = dayNumber
    }
  }
}
```

----

~~Unfortunately, we need an API change here. The `.calendarItemModel` extension on `SwiftUI.View` needs to take in an `AnyHashable` ID, which we internally use to give the view a stable identifier, even when reused across hosting controllers. For simple cases, one can simply pass a `Day` or a `Month` (both of which are `Hashable`) as the identifier.~~

IDs are set internally, so this is a non-breaking change.

We might also consider providing an API that looks like this in the future:

```swift
CalendarViewRepresentable(...)
  .dayView { dayComponents in
    Text(dayComponents.day)
  }
```

This closure is a view builder, and leaves the conversion to a `CalendarItemModel<SwiftUIWrapperView<SwiftUI.View>>` as an implementation detail (which also means we would handle applying IDs to views internally). I haven't fully thought through if this is possible, but it'd be a nice improvement.

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/256

## Motivation and Context

Bug fix, make SwiftUI integration better

## How Has This Been Tested

Example app

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/AnyCalendarItemModel.swift
+++ b/Sources/Public/AnyCalendarItemModel.swift
@@ -43,6 +43,8 @@ public protocol AnyCalendarItemModel {
   /// - Note: There is no reason to invoke this function from your feature code; it should only be invoked internally.
   func _isContentEqual(toContentOf other: AnyCalendarItemModel) -> Bool
 
+  mutating func _setSwiftUIWrapperViewContentIDIfNeeded(_ id: AnyHashable)
+
 }
 
 // MARK: - _CalendarItemViewDifferentiator
--- a/Sources/Public/ItemViews/SwiftUIWrapperView.swift
+++ b/Sources/Public/ItemViews/SwiftUIWrapperView.swift
@@ -30,8 +30,10 @@ public final class SwiftUIWrapperView<Content: View>: UIView {
 
   // MARK: Lifecycle
 
-  public init(content: Content) {
-    self.content = content
+  public init(contentAndID: ContentAndID) {
+    self.contentAndID = contentAndID
+    hostingController = HostingController(
+      rootView: .init(content: contentAndID.content, id: contentAndID.id))
 
     super.init(frame: .zero)
 
@@ -69,16 +71,16 @@ public final class SwiftUIWrapperView<Content: View>: UIView {
 
   // MARK: Fileprivate
 
-  fileprivate var content: Content {
+  fileprivate var contentAndID: ContentAndID {
     didSet {
-      hostingController.rootView = content
+      hostingController.rootView = .init(content: contentAndID.content, id: contentAndID.id)
       configureGestureRecognizers()
     }
   }
 
   // MARK: Private
 
-  private lazy var hostingController = HostingController<Content>(rootView: content)
+  private let hostingController: HostingController<IDWrapperView<Content>>
 
   private weak var hostingControllerView: UIView?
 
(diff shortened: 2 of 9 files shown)


bryankeller:
@bryn no rush on a review - I'm actually going to revise this a bit to see if I can make this API change purely additive. I think there's a way to manage these IDs internally so that consumers don't need to specify them in their item provider closures.

Sources/Public/AnyCalendarItemModel.swift
```diff
@@ -43,6 +43,8 @@ public protocol AnyCalendarItemModel {
   /// - Note: There is no reason to invoke this function from your feature code; it should only be invoked internally.
   func _isContentEqual(toContentOf other: AnyCalendarItemModel) -> Bool
 
+  mutating func _setSwiftUIWrapperViewContentIDIfNeeded(_ id: AnyHashable)
```
bryankeller, line 46: kind of annoying that we need this but we're working with a type-erased AnyCalendarItemModel, so we don't know if the original `CalendarItemModel<CalendarItemViewRepresentable>` type is a `SwiftUIWrapperView` or not.

Sources/Public/ItemViews/SwiftUIWrapperView.swift
```diff
@@ -188,35 +204,48 @@ extension UIResponder {
...
+  @MainActor required dynamic init?(coder aDecoder: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  override func viewDidLoad() {
+    super.viewDidLoad()
 
+    // Override the default `.systemBackground` color since `CalendarView` subviews should be
+    // clear.
+    view.backgroundColor = .clear
```
bryankeller, line 248: unchanged, but I pulled it out of the `SwiftUIWrapperView` extension since it was private anyways.

Sources/Public/ItemViews/SwiftUIWrapperView.swift
```diff
@@ -188,35 +204,48 @@ extension UIResponder {
...
 
-  /// The `UIHostingController` type used by `SwiftUIWrapperView` to embed SwiftUI views in a UIKit view hierarchy.
-  public final class HostingController<Content: View>: UIHostingController<Content> {
+  let content: Content
+  let id: AnyHashable
 
-    // MARK: Lifecycle
+  var body: some View {
+    content
+      .id(id)
```
bryankeller, line 218: the magic line that adds a stable ID

brynbodayle (approved):

PR #269, opened 2023-09-18
Update CalendarViewContent closures to accept nil return values

calda:
This PR updates the closures on `CalendarViewContent` to accept nil return values. This makes it easier for consumers to handle unexpected cases. 

For example, today using `[weak self]` with `CalendarViewContent` is tricky:

```swift
CalendarViewContent(
  visibleDateRange: Date.distantPast...Date.distantFuture, 
  monthsLayout: .vertical)
  .dayItemProvider { [weak self] day in
    guard let self else { 
      // We have to return a non-optional value, but can't.
      // The only option is to crash.
      preconditionFailure("self unexpectedly deallocated already") 
    }
        
    return self.dayItemProvider(day)
  }
```

after these changes its much easier:

```swift
CalendarViewContent(
  visibleDateRange: Date.distantPast...Date.distantFuture, 
  monthsLayout: .vertical)
  .dayItemProvider { [weak self] day in
    guard let self else { return nil }
    return self.dayItemProvider(day)
  }
```

### DIFF
--- a/Sources/Internal/VisibleItemsProvider.swift
+++ b/Sources/Internal/VisibleItemsProvider.swift
@@ -950,11 +950,13 @@ final class VisibleItemsProvider {
       daysAndFrames: dayRangeLayoutContext.daysAndFrames,
       boundingUnionRectOfDayFrames: dayRangeLayoutContext.boundingUnionRectOfDayFrames)
 
-    context.visibleItems.insert(
-      VisibleItem(
-        calendarItemModel: dayRangeItemProvider(dayRangeLayoutContext),
-        itemType: .dayRange(dayRange),
-        frame: frame))
+    if let dayRangeItemModel = dayRangeItemProvider(dayRangeLayoutContext) {
+      context.visibleItems.insert(
+        VisibleItem(
+          calendarItemModel: dayRangeItemModel,
+          itemType: .dayRange(dayRange),
+          frame: frame))
+    }
   }
 
   private func handlePinnedDaysOfWeekIfNeeded(
@@ -1029,7 +1031,8 @@ final class VisibleItemsProvider {
         let layoutContext = overlayLayoutContext(
           for: overlaidItemLocation,
           inBounds: bounds,
-          context: &context)
+          context: &context),
+        let itemModel = itemModelProvider(layoutContext)
       else
       {
         continue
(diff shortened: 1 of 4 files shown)


Sources/Public/CalendarViewContent.swift
```diff
@@ -208,19 +175,27 @@ public final class CalendarViewContent {
...
   public func monthHeaderItemProvider(
-    _ monthHeaderItemProvider: @escaping (_ month: Month) -> AnyCalendarItemModel)
+    _ monthHeaderItemProvider: @escaping (_ month: Month) -> AnyCalendarItemModel?)
     -> CalendarViewContent
   {
-    self.monthHeaderItemProvider = monthHeaderItemProvider
+    self.monthHeaderItemProvider = { [defaultMonthHeaderItemProvider] month in
+      guard let itemModel = monthHeaderItemProvider(month) else {
+        // If the caller returned nil, fall back to the default item provider
+        return defaultMonthHeaderItemProvider(month)
```
calda, line 193: To avoid having to change any of the existing call sites, we can have the instance properties still always return a non-nil value, even if the user-provided closure returns nil. An easy way to do this is to just fall back to the existing default item provider, which always returns a valid value.

Sources/Internal/VisibleItemsProvider.swift
```diff
@@ -950,11 +950,13 @@ final class VisibleItemsProvider {
       daysAndFrames: dayRangeLayoutContext.daysAndFrames,
       boundingUnionRectOfDayFrames: dayRangeLayoutContext.boundingUnionRectOfDayFrames)
 
-    context.visibleItems.insert(
-      VisibleItem(
-        calendarItemModel: dayRangeItemProvider(dayRangeLayoutContext),
-        itemType: .dayRange(dayRange),
-        frame: frame))
+    if let dayRangeItemModel = dayRangeItemProvider(dayRangeLayoutContext) {
```
calda, line 953: `dayRangeItemProvider` and `overlaidItemLocationsAndItemProvider` are already optional on `CalendarViewContent`, so we can already gracefully handle them being nil.

bryankeller: This creates a bit of an odd scenario. The way those 2 APIs work is that the first parameter is a set of date ranges / overlay locations that we're saying _will have_ a corresponding view. It really shouldn't be possible to return `nil`.

For example, if we do

```swift
CalendarViewContent(...)
  .dayRangeItemProvider(for: [selectedDateRange]) { ... }
```

then we really shouldn't be allowed to return `nil` from that closure. If we want to omit a day range, then we should omit it from that set of date ranges that we pass in as the first param.

bryankeller (approved):
Awesome, thanks @calda !

PR #272, opened 2023-09-22
Bk/performance improvements

bryankeller:
## Details

This PR makes 2 performance improvements:

- Prevent a few `DateFormatter`s from getting recreated on every `CalendarViewContent` creation
- Prevent `updateVisibleViews(...)` from getting called if the calendar hasn't scrolled far enough for the visible set of views to have changed

## Related Issue

N/A

## Motivation and Context

Optimizing host calendar in the Airbnb app

## How Has This Been Tested

Example app, profiling.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -258,6 +258,10 @@ public final class CalendarView: UIView {
       invalidateIntrinsicContentSize()
     }
 
+    // Clear this so that we don't return early in our next layout pass even though our layout
+    // region might not have changed.
+    layoutRegion = nil
+
     self.content = content
     setNeedsLayout()
 
@@ -496,6 +500,7 @@ public final class CalendarView: UIView {
   private var anchorLayoutItem: LayoutItem?
   private var _visibleItemsProvider: VisibleItemsProvider?
   private var visibleItemsDetails: VisibleItemsDetails?
+  private var layoutRegion: ClosedRange<LayoutItem.ItemType>?
   private var visibleViewsForVisibleItems = [VisibleItem: ItemView]()
 
   private var isAnimatedUpdatePass = false
@@ -734,11 +739,14 @@ public final class CalendarView: UIView {
       isAnimatedUpdatePass: isAnimatedUpdatePass)
     self.anchorLayoutItem = currentVisibleItemsDetails.centermostLayoutItem
 
-    updateVisibleViews(
-      withVisibleItems: currentVisibleItemsDetails.visibleItems,
-      previouslyVisibleItems: visibleItemsDetails?.visibleItems ?? [])
+    // If our first / last layout item hasn't changed, then we haven't scrolled enough to trigger
+    // an update of visible views. This short-circuit greatly improves scroll performance.
+    if currentVisibleItemsDetails.layoutRegion != layoutRegion {
+      updateVisibleViews(withVisibleItems: currentVisibleItemsDetails.visibleItems)
+    }
 
     visibleItemsDetails = currentVisibleItemsDetails
+    layoutRegion = currentVisibleItemsDetails.layoutRegion
 
     let minimumScrollOffset = visibleItemsDetails?.contentStartBoundary.map {
       ($0 - firstLayoutMarginValue).alignedToPixel(forScreenWithScale: scale)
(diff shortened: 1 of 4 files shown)


Sources/Public/CalendarViewContent.swift
```diff
@@ -407,22 +392,20 @@ public final class CalendarViewContent {
...
-  ] month in
-    let firstDateInMonth = calendar.firstDate(of: month)
-    let monthText = monthHeaderDateFormatter.string(from: firstDateInMonth)
-    let itemModel = MonthHeaderView.calendarItemModel(
-      invariantViewProperties: .base,
-      content: .init(monthText: monthText, accessibilityLabel: monthText))
-    return itemModel
-  }
+  private lazy var defaultMonthHeaderItemProvider: (Month) -> AnyCalendarItemModel = 
+    { [unowned self] month in
```
bryankeller, line 396: Should be impossible to hit cc @calda

Sources/Public/CalendarView.swift
```diff
@@ -258,6 +258,10 @@ public final class CalendarView: UIView {
       invalidateIntrinsicContentSize()
     }
 
+    // Clear this so that we don't return early in our next layout pass even though our layout
+    // region might not have changed.
+    layoutRegion = nil
```
bryankeller, line 263: we clear this in `setContent`, since we want `updateVisibleViews(...)` to run on the next layout pass even if our layout region hasn't changed

Sources/Public/CalendarView.swift
```diff
@@ -734,11 +739,14 @@ public final class CalendarView: UIView {
...
-      withVisibleItems: currentVisibleItemsDetails.visibleItems,
-      previouslyVisibleItems: visibleItemsDetails?.visibleItems ?? [])
+    // If our first / last layout item hasn't changed, then we haven't scrolled enough to trigger
+    // an update of visible views. This short-circuit greatly improves scroll performance.
+    if currentVisibleItemsDetails.layoutRegion != layoutRegion {
+      updateVisibleViews(withVisibleItems: currentVisibleItemsDetails.visibleItems)
+    }
 
     visibleItemsDetails = currentVisibleItemsDetails
+    layoutRegion = currentVisibleItemsDetails.layoutRegion
```
bryankeller, line 749: here's the big optimization - when we're scrolling, we only need to `updateVisibleViews(...)` if we've scrolled far enough to change our "layout region" (month headers, days, etc. enter / exit the viewport)

Sources/Public/CalendarViewContent.swift
calda: What do you mean?

bryankeller: @calda I mean the unowned should be "safe" / there's no way to crash here.

calda: Perhaps -- although not sure I understand the motivation to change from capturing individual properties (guaranteed to be safe) to using `unowned` instead (potentially unsafe).

You could feasibly construct a case that crashes by doing something like this:

```swift
let content = CalendarViewContent(...)

let monthHeaderItemProvider = content.monthHeaderItemProvider

/// ... later, in a different scope, so content itself has been deallocated

monthBackgroundItemProvider(month) // 💥 crash because content no longer exists but there was an unowned reference to it
```

so the safety depends on the usage of the `content.monthHeaderItemProvider` property

brynbodayle (approved):

Sources/Public/CalendarViewContent.swift
bryankeller: That would only be possible internally though, since CalendarViewContent doesn't expose its item provider closures publicly.

Capturing the individual properties here causes them to be created immediately (which is expensive for date formatters).

```swift
lazy var expensiveProperty = ExpensiveObject()

let foo = { [expensiveProperty] in 
  expensiveProperty.doSomething()
}

// later...
foo()
```

In this example, `expensiveProperty` is initialized when `foo` is initialized, _not_ when `foo` is called later on.

calda: That makes sense -- I guess in a lot of typical usage we'd be initializing the `monthHeaderDateFormatter` for the `defaultMonthHeaderItemProvider` but then never even calling the default provider (because the consumer immediately provides a custom provider)

PR #283, opened 2023-11-18
Accessibility scrolling enhancements

andrewwilliams:
## Details

When scrolling the `CalendarView` via accessibility methods (specifically using the 3 finger swipe with voiceover), the related scrolling delegate methods (`didScroll`, `didEndDragging`, and `didEndDecelerating`) were not called. 

This could lead to missing functionality for surfaces which relied on those callbacks to implement functionality, such as loading new data from the network when scrolling to new a new data range.

This PR adds those callbacks to the `accessibilityScroll` method, ensuring that they're fired whenever a user scrolls - whether through accessibility means or not.

## How Has This Been Tested

Tested changes in example app on device to ensure that the relevant delegate methods are called.

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -1259,6 +1259,13 @@ extension CalendarView {
     let accessibilityScrollText = targetMonthView.accessibilityLabel
     UIAccessibility.post(notification: .pageScrolled, argument: accessibilityScrollText)
 
+    // ensure that scrolling related callbacks are still fired when performing scrolling via accessibility
+    if let visibleDayRange {
+      didEndDecelerating?(visibleDayRange)
+      didEndDragging?(visibleDayRange, false)
+      didScroll?(visibleDayRange, true)
+    }
+
     return true
   }
 
--- a/Sources/Public/CalendarView.swift
+++ b/Sources/Public/CalendarView.swift
@@ -1261,9 +1261,9 @@ extension CalendarView {
 
     // ensure that scrolling related callbacks are still fired when performing scrolling via accessibility
     if let visibleDayRange {
+      didScroll?(visibleDayRange, false)
+      didEndDragging?(visibleDayRange, true)
       didEndDecelerating?(visibleDayRange)
-      didEndDragging?(visibleDayRange, false)
-      didScroll?(visibleDayRange, true)
     }
 
     return true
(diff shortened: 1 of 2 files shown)


jpsoultanis (approved):
🎉

Sources/Public/CalendarView.swift
```diff
@@ -1259,6 +1259,13 @@ extension CalendarView {
     let accessibilityScrollText = targetMonthView.accessibilityLabel
     UIAccessibility.post(notification: .pageScrolled, argument: accessibilityScrollText)
 
+    // ensure that scrolling related callbacks are still fired when performing scrolling via accessibility
+    if let visibleDayRange {
+      didEndDecelerating?(visibleDayRange)
+      didEndDragging?(visibleDayRange, false)
+      didScroll?(visibleDayRange, true)
```
bryankeller, line 1266: isn't this order flipped? I would expect this order, and these parameters:

```suggestion
      didScroll?(visibleDayRange, false)
      didEndDragging?(visibleDayRange, true)
      didEndDecelerating?(visibleDayRange)      
```

bryankeller:
Thanks Andrew! Left one comment / question. Also, could you update the `CHANGELOG.md`? This can be under the "Fixed" section.

Sources/Public/CalendarView.swift
andrewwilliams: Yup, they are in the exact opposite order they should be in 🤦 . Thanks for catching that!

For the parameters, I thought we might want to set didScroll / isDragging to true since the action was performed by the user via voiceover? But maybe I'm misunderstanding how that parameter should be used.

For `didEndDragging`, I thought we would want to set `willDecelerate` to false since after this is fired there will be no further scrolling/deceleration?

bryankeller: > For the parameters, I thought we might want to set didScroll / isDragging to true since the action was performed by the user via voiceover? But maybe I'm misunderstanding how that parameter should be used.

I think that logic makes sense actually, let's leave that one as true

> For didEndDragging, I thought we would want to set willDecelerate to false since after this is fired there will be no further scrolling/deceleration?

This one is a bit tricky - the accessibility scroll has an animation / what appears to be some deceleration period, right? I think we should just do whatever collection view does for these 2 calls.

andrewwilliams: Confirmed collection view behavior - it was exactly as you suggested. Addressed via fixup.

bryankeller (approved):

PR #284, opened 2023-11-28
Simplify and improve accessibility implementation

bryankeller:
## Details

This simplifies and also improves support for Voice Over. Previously, we had the concept of on-screen accessibility elements, and some special off-screen accessibility elements that we used to make off-screen content focusable by Voice Over to prevent the dreaded "heading not found" issue that happens with lazy views like this and `UICollectionView`.

That approach had some issues with SwiftUI views, since it made some assumptions about the off-screen accessibility element being based on a `UIView`.

I rethought things a bit, and realized we could simplify. I had recently added support for doing an extended bounds layout pass, which forces off-screen views to be created and laid out for the purpose of having animations look better. We can reuse this code when Voice Over is running to ensure that nearby off-screen views can be focused. This change allows us to unify how all accessibility elements are treated, no longer needing to differentiate between off-screen and on-screen elements.

## Related Issue

N/A

## Motivation and Context

Fixes accessibility behavior for SwiftUI views

## How Has This Been Tested

Example app

## Types of changes

## Checklist


Sources/Internal/CalendarScrollView.swift
```diff
@@ -0,0 +1,81 @@
...
+  required init?(coder _: NSCoder) {
+    fatalError("init(coder:) has not been implemented")
+  }
+
+  // MARK: Internal
+
+  var cachedAccessibilityElements: [Any]?
+
+  override var contentInsetAdjustmentBehavior: ContentInsetAdjustmentBehavior {
+    didSet {
```
brynbodayle, line 42: is it safer to do get/set like we do below? wondering if we could get into loops more easily this way

brynbodayle (approved):

PR #297, opened 2024-01-26
Bk/another accessibility refactor

bryankeller:
## Details

This PR makes further accessibility improvements. The main thing here is that we no longer post `.layoutChanged` or `.screenChanged` notifications to force focus, and instead rely on UIScrollView to handle maintaining element focus for us (which works well, as long as views aren't reused!).

Speaking of that, this disables view recycling when voice over is active. This greatly simplifies things, and removes the need to constantly try to regain control over the focus. I should have thought of this sooner 🫠 We still have flat memory usage since off-screen views are simply deallocated. We pay a small CPU performance price, but Voice Over users won't be scrolling super fast anyways, so the tradeoff seems good. The overall UX for Voice Over users is significantly improved since focus never temporarily jumps around.

## Related Issue

N/A

## Motivation and Context

Improve VO UX

## Types of changes

## Checklist

### DIFF
--- a/Sources/Internal/ItemViewReuseManager.swift
+++ b/Sources/Internal/ItemViewReuseManager.swift
@@ -25,6 +25,7 @@ final class ItemViewReuseManager {
 
   func viewsForVisibleItems(
     _ visibleItems: Set<VisibleItem>,
+    recycleOldViews: Bool,
     viewHandler: (
       ItemView,
       VisibleItem,
@@ -56,6 +57,7 @@ final class ItemViewReuseManager {
 
       let context = reusedViewContext(
         for: visibleItem,
+        recycleOldViews: recycleOldViews,
         unusedPreviouslyVisibleItems: &visibleItemsDifference)
       viewHandler(
         context.view,
@@ -76,6 +78,7 @@ final class ItemViewReuseManager {
 
   private func reusedViewContext(
     for visibleItem: VisibleItem,
+    recycleOldViews: Bool,
     unusedPreviouslyVisibleItems: inout Set<VisibleItem>)
     -> ReusedViewContext
   {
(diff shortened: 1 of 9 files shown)


Sources/Internal/ItemViewReuseManager.swift
```diff
@@ -25,6 +25,7 @@ final class ItemViewReuseManager {
 
   func viewsForVisibleItems(
     _ visibleItems: Set<VisibleItem>,
+    recycleOldViews: Bool,
```
brynbodayle, line 28: Something more descriptive?
```suggestion
    recycleOffscreenViews: Bool,
```

Sources/Public/CalendarView.swift
```diff
@@ -526,6 +520,14 @@ public final class CalendarView: UIView {
   private lazy var scrollViewDelegate = ScrollViewDelegate(calendarView: self)
   private lazy var gestureRecognizerDelegate = GestureRecognizerDelegate(calendarView: self)
 
+  private var hasHadInitialItemViewFocus = false {
```
brynbodayle, line 523:
```suggestion
  private var initialItemViewWasFocused = false {
```

brynbodayle (approved):
Good trade off!

Sources/Internal/ItemViewReuseManager.swift
bryankeller: Nice, I like that

bryankeller: They might not be offscreen, so going with `recycleUnusedViews` (imagine we just switch from using `DayViewA` to `DayViewB` for a visible day, it's not really offscreen in terms of being within the visible bounds of the scroll view)

Issue #230, opened 2023-04-30
programmatically update de selection range 


alexisgarciaa:
Hi @bryankeller thank you so much for this amazing library i have been trying to figure out how update the selection range using a a button outside of the calendar but i got that the model Day is private so we can not create a custom day and i saw that you explained why is like that but i would love to hear if this is possible using another method, thank you so much in advance by the way i am using SwiftUI

bryankeller:
Hey @alexisgarciaa - can you post some example code so I can understand what you've tried?

PR #298, opened 2024-01-27
Make DayComponentRange initializer public to support pre-selected date range in the calendar.

IvanUshakov:
## Details

Make initializer of DayComponentRange with Date range public

## Related Issue

[programmatically update de selection range](https://github.com/airbnb/HorizonCalendar/issues/230)

## Motivation and Context

In the application that I am currently developing there is a need for a calendar with a selection of date ranges. In this case, I need to pass the date range already selected by the user to the calendar itself, and he can edit it. In your example you use DayComponentsRange to represent date ranges, this is very convenient and works for me. But I can't initialize it with Date objects. So I'll just make initializer public

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/DayRange.swift
+++ b/Sources/Public/DayRange.swift
@@ -27,7 +27,7 @@ extension DayComponentsRange {
 
   /// Instantiates a `DayRange` that encapsulates the `dateRange` in the `calendar` as closely as possible. For example,
   /// a date range of [2020-05-20T23:59:59, 2021-01-01T00:00:00] will result in a day range of [2020-05-20, 2021-01-01].
-  init(containing dateRange: ClosedRange<Date>, in calendar: Calendar) {
+  public init(containing dateRange: ClosedRange<Date>, in calendar: Calendar) {
     self.init(
       uncheckedBounds: (
         lower: calendar.day(containing: dateRange.lowerBound),


bryankeller:
Hi @IvanUshakov, there's no need to create this type yourself - it's initialized is intentionally private.

What exactly is your use case? There are no parts of the public API of HorizonCalendar that require you to pass in these types, so I'd like to better understand what you're trying to do.

My recommendation is to convert DayComponents and DayRangeComponents to Foundation Dates, using the correct Foundation Calendar instance.

IvanUshakov:
In the example there is a class DayRangeSelectionTracker which shows well how you can select a range of dates. SwiftUIScreenDemo and DayRangeSelectionTracker uses DayComponentsRange for this. I need to pass the current selected dates by the user to a class similar to SwiftUIScreenDemo, but I cannot do because the initializer is declared as internal. Of course I can use system types for this, but it's not as easy as with DayComponentsRange. And if the user has to use system types, then perhaps the example should also use system types. Because in a large number of cases the user will edit the selected dates, and not just select them from scratch.

Сode is not correct, but it shows the idea:
```swift
struct AppCalendarView: View {
    @State private var selectedDayRange: DayComponentsRange?
    @Binding private var selectedRange: ClosedRange<Date>?

    init(selectedRange: Binding<ClosedRange<Date>?>) {
        _selectedRange = selectedRange
        if let selectedRangeBinding = selectedRangeBinding?.wrappedValue {
            self.selectedDayRange = .init(containing: selectedRangeBinding, in: calendar)
        }
        ...
    }
    ...
}
```

IvanUshakov:
@bryankeller have you any updates on this?

canyoldas0:
> Hi @IvanUshakov, there's no need to create this type yourself - its initializer is intentionally private.

I share the same motivation on this. #300 would be nice to be more flexible about the initializers, can't really follow up why those initializers are private intentionally.

varyamereon:
@bryankeller Was there an issue with this pull request? Seems it's been waiting to merge for a while and this would be very useful for me!!

Issue #299, opened 2024-02-03
Laggy scroll when using custom day view in `CalendarViewRepresentable`


kevalkorat:
CalendarView's scroll doesn't seem smooth when adding a custom day view. Currently trying to use this calendar for a fitness app I am making where I want to show the yearly view of the calendar along with some data for each day. I have also tried to split the `CalendarDayView` even further but that didn't solve the scroll issue. 

This is my setup and you can run on device and see the performance issue. Scroll is just not silky smooth and seems janky.

Can someone help to see if I am doing something wrong here and maybe fixing that can improve the scroll. Any help is greatly appreciated. @bryankeller 

```swift

struct HorizonCalendarView: View {
    let startDate = Calendar.current.date(from: DateComponents(year: 2020, month: 01, day: 01))!
    let endDate = Date.now
    
    var body: some View {
        CalendarViewRepresentable(
            visibleDateRange: startDate...endDate,
            monthsLayout: .vertical,
            dataDependency: nil)
        
        .verticalDayMargin(4)
        .dayAspectRatio(1.25)
        .interMonthSpacing(30)
        .dayOfWeekAspectRatio(0.5)
        .monthDayInsets(.init(EdgeInsets(top: -20, leading: 0, bottom: 0, trailing: 0)))
        
        .dayOfWeekHeaders({ month, weekdayIndex in
            EmptyView()
        })
        
        .dayItemProvider { day in
            CalendarDayView(dayNumber: day.day, isSelected: false, reps: 100)
            .calendarItemModel
        }
    }
}

struct CalendarDayView: View {
    let dayNumber: Int
    let isSelected: Bool
    var reps: Int?
    
    var repsString: String {
        guard let reps else {
            return "0"
        }
        return "\(reps)"
    }
    
    var dayTextColor: Color {
        isSelected ?
            .primary : .init(uiColor: .tertiaryLabel)
    }
    
    var repsTextColor: Color {
        switch reps {
        case .none: return .gray.opacity(0.15)
        case .some: return Color(uiColor: .systemBackground)
        }
    }
    
    var body: some View {
            VStack(spacing: 0) {
                Text("\(dayNumber)")
                    .foregroundStyle(dayTextColor)
                    .padding(.init(top: 4, leading: 7, bottom: 4, trailing: 7))
                    .frame(maxWidth: .infinity)
                    .font(.caption)
                
                Text(repsString)
                    .foregroundStyle(repsTextColor)
[...]

PR #320, opened 2024-11-09
Optimize SwiftUIWrapperView

bryankeller:
## Details

This PR significantly reduces CPU use for people using SwiftUI views for their calendar items by making 2 optimizations:
1. Don't add the `UIHostingController` as a child view controller, and just work with its hosting view directly
2. Don't give every view a unique ID just to get `onAppear` and `onDisappear` callbacks - instead, set a hidden view's to just display an `EmptyView`

Together, these changes reduce CPU use by 35-40% when scrolling fast.

## Related Issue

https://github.com/airbnb/HorizonCalendar/issues/299

## Motivation and Context

Performance improvements.

## How Has This Been Tested

Example app

## Types of changes

## Checklist

### DIFF
--- a/Sources/Public/ItemViews/SwiftUIWrapperView.swift
+++ b/Sources/Public/ItemViews/SwiftUIWrapperView.swift
@@ -22,9 +22,6 @@ import SwiftUI
 /// Consider using the `calendarItemModel` property, defined as an extension on SwiftUI's`View`, to avoid needing to work with
 /// this wrapper view directly.
 /// e.g. `Text("\(dayNumber)").calendarItemModel`
-///
-/// - Warning: Using a SwiftUI view with the calendar will cause `SwiftUIView.HostingController`(s) to be added to the
-/// closest view controller in the responder chain in relation to the `CalendarView`.
 @available(iOS 13.0, *)
 public final class SwiftUIWrapperView<Content: View>: UIView {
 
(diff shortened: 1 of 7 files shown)


Sources/Public/ItemViews/SwiftUIWrapperView.swift
```diff
@@ -47,16 +47,20 @@ public final class SwiftUIWrapperView<Content: View>: UIView {
...
 
-  public override func didMoveToWindow() {
-    super.didMoveToWindow()
-
-    if window != nil {
-      setUpHostingControllerIfNeeded()
+  public override var isHidden: Bool {
+    didSet {
+      if isHidden {
+        hostingController.rootView = AnyView(EmptyView())
```
bryankeller, line 62: not ideal, but way faster than the previous strategy of giving every view an ID

bryankeller: We could also add an API in the future to opt-out of this behavior if someone doesn't need `onAppear` or `onDisappear` callbacks in their views, which would get them back ~20% CPU usage when scrolling fast 🤯

Sources/Public/ItemViews/SwiftUIWrapperView.swift
```diff
@@ -76,62 +80,42 @@ public final class SwiftUIWrapperView<Content: View>: UIView {
...
+      withHorizontalFittingPriority: horizontalFittingPriority,
+      verticalFittingPriority: verticalFittingPriority)
   }
 
   // MARK: Fileprivate
 
   fileprivate var contentAndID: ContentAndID {
     didSet {
-      hostingController.rootView = .init(content: contentAndID.content, id: contentAndID.id)
+      hostingController.rootView = AnyView(contentAndID.content)
```
bryankeller, line 102: I profiled and there was no perf hit from switching to `AnyView`

PR #321, opened 2024-11-14
Fix sizing month header on every frame and add more calendar model caching

bryankeller:
## Details

This PR fixes 2 issues:

1. I was never caching the result of self-sizing a month header, causing us to call systemLayoutSizeFitting on every frame 💀
2. We weren't caching calendar item models for some calendar slots, like for day ranges and month backgrounds

## Related Issue

N/A

## Motivation and Context

Performance improvements

## How Has This Been Tested

Example app, Airbnb app


brynbodayle (approved):
Would be nice to add some tests for this

bryankeller:
Hmm yeah, I'll think about how that might be possible. We don't expose whether something was returned from cache or not, but I bet there's some way we could get it tested.

PR #334, opened 2025-05-07
Make DayComponents initializer public for external use

Rk1821389:
## Details

This change updates the access control of the `DayComponents` initializer from `internal` to `public`. This enables developers to instantiate `DayComponents` directly when using the package via Swift Package Manager, which is currently impossible due to the restricted access level.

```swift
// Before
init(month: MonthComponents, day: Int) { ... }

// After
public init(month: MonthComponents, day: Int) { ... }

### DIFF
--- a/Sources/Public/Day.swift
+++ b/Sources/Public/Day.swift
@@ -28,7 +28,7 @@ public struct DayComponents: Hashable {
 
   // MARK: Lifecycle
 
-  init(month: MonthComponents, day: Int) {
+  public init(month: MonthComponents, day: Int) {
     self.month = month
     self.day = day
   }


keepfocused:
Please @bryankeller @brynbodayle 
I need this soo much. This limits my capabilities to work with library a lot. 
I cannot initialise variable of type DayComponents or MonthComponents
Screenshot 2025-10-17 at 13 58 11

bryankeller:
`DayComponents` and `MonthComponents` can be thought of as equivalent to `DateComponents`, which you are welcome to make yourself since they are part of `Foundation`. There is no need to create `DayComponents` or `MonthComponents`, as there is no API that expects you to provide those to the calendar. Please use a `Calendar` instance to create `Date`s in the correct timezone. This approach ensures that you do not introduce timezone-related bugs.

